PR #16, opened 2015-01-13
Adds subdomain validation

matthopson:


semmons99:
appears solid to me :rocket:

PR #17, opened 2015-01-27
Remove events in repeating event series after saveEvent

kristenboydstun:
If the date range for a repeating event changes, some events in the
series may be deleted. Check for deleted events when a repeating
event is updated and unlink them.


matthopson:
:+1:

PR #21, opened 2015-03-06
Add weather forecast support for first-party products.

matthopson:


knappe:
Matt and I spoke about this PR and I think we're going to hold off on this until a few concerns are addressed.

matthopson:
Killing this off. :)

PR #23, opened 2015-03-12
Add notifyTeamAsMemberId to event requests

kristenboydstun:
Event creation/update/deletion expects the `notifyTeamAsMemberId` param when `notifyTeam` is true.


matthopson:
Let's discuss a bit before proceeding. :)

matthopson:
:( :+1:

PR #32, opened 2015-04-20
Latest Brunch was building incorrectly - temp fix until resolved.

matthopson:
This is a temporary fix to force an older version of Brunch while an issue is sorted out with compiling with latest brunch.


kristenboydstun:
:+1:

PR #34, opened 2015-04-22
Remove team files

matthopson:
Removes support for `teamFiles` as this is being refactored in APIv3 to better support Media in TeamSnap. This was initially added to the SDK as a means to track whether a team has files / photos to export, but this has now been addressed with `team.has_exportable_media` in APIv3.


glanotte:
looks good :+1:

PR #41, opened 2015-05-27
Bump version to 1.2.0

matthopson:
We've been less than responsible in terms of versioning. That ends now.


kristenboydstun:
:clap: :ship:

PR #44, opened 2015-06-28
Removes payments associated with deleted member

matthopson:
When a user is deleted, it should also unlink associated payments or suffer the wrath of javascript errors and the gnashing of teeth.


semmons99:
Tests?

matthopson:
We haven't really ever written tests for this feature since it essentially
unlinks the data in memory - it isn't deleting anything as that's happening
in API automagically.

Think of it as a cache-buster.

matthopson:
That said - I could probably work up some tests for these if you think we
should have coverage on it.

semmons99:
I’ll let you be the judge. So long as you feel the SDK is sufficiently
exercised.


*Shane Emmons* | Director of Technology and Principal Architect
TeamSnap - The easiest way to manage your team
(810) 834-9542 | shane@teamsnap.com | What is TeamSnap?
<http://www.youtube.com/watch?v=fx0vRXYGijU>

matthopson:
@semmons99 Chatting with @kristenboydstun, a bit about this... She shares a similar opinion for this case. It's probably less beneficial to write tests over persistence in individual items - as this uses a core SDK method: linking and unlinking items... and that as long as unlinking is doing its job, we shouldn't see failures here. Linking and unlinking being one of the more exercised features of the SDK, we'd have bigger problems if this was failing (and would certainly know right away).

semmons99:
excellent. :shipit:

PR #45, opened 2015-07-07
add division standings

dmccraw:
I haven't built teamsnap.js since it makes too many changes when I try to build.


matthopson:
:+1:  I'll build and release with my stats PR

PR #46, opened 2015-07-08
Load payments on member save

matthopson:
Fix bug where saving member didn't automatically load `memberPayments`.


semmons99:
:shipit:

PR #47, opened 2015-07-14
Statistics

matthopson:
@kristenboydstun as promised :smiley:


jjmocko:
This being my first delve into the sdk, it looks good. Do you need testing around the memberStatistic and teamStatistic endpoints as well, or is that covered in the other added tests? Someone else who has more experience with the sdk should sign off, but :+1:

matthopson:
Thanks @jjmocko! Good catch! I've added some tests to cover those.

semmons99:
Looks legit to me :shipit:

PR #48, opened 2015-07-17
remove divisionTeamStanding as a teamType

dmccraw:


glanotte:
seems right to me :+1:

matthopson:
So this works if we want to pull it from the SDK - but it seems we probably want to pull it from NJ instead... or are you not going to be bulk loading this later @dmccraw?

dmccraw:
I think we should pull it from the bulk_loader for now until we figure out what we want to do with all the division/league information.

dmccraw:
Moving this change to nextjenn

PR #50, opened 2015-08-06
Fix sort

matthopson:
- Does a better job of sorting names with accented characters
- Takes care of some lint warnings that have been hanging around


semmons99:
:metal:

Issue #31, opened 2015-04-07
request is defined twice


ORESoftware:
was looking through the code (looks good), but one minor issue is that request is defined twice in teamsnap.js

request = require('./request');

promises = require('./promises');

request = require('./request');

(you'll see it)

falonofthetower:
Possibly resolved with this https://github.com/teamsnap/teamsnap-javascript-sdk/pull/57

Sorry about the multiple attempts was trying to properly format the commit from the command line.

PR #57, opened 2015-09-14
Remove second request = require...

falonofthetower:
Per issue #31 on git request is defined twice, this just removes the second instance

### DIFF
--- a/lib/teamsnap.js
+++ b/lib/teamsnap.js
@@ -123,8 +123,6 @@ request = require('./request');
 
 promises = require('./promises');
 
-request = require('./request');
-
 jsonMime = 'application/json';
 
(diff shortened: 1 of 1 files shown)


matthopson:
Hi @falonofthetower, this issue will actually be addressed in an upcoming release, however, I'd like to point out that `teamsnap.js` is actually built from the `/src` directory with brunch, so any changes to this js file would just be overwritten.

Honestly, this should be addressed in a "contributing" section on the wiki - which is yet to exist - so my apologies there. If you'd like to create a new issue for the contributing section to track it, you're more than welcome.

Also I've enabled Gitter for this repo so folks having issues or questions can post there.

Thanks!

PR #63, opened 2015-09-25
Team media

matthopson:
- Support for Media Tab
- Fixes bug with upload progress handlers


src/collections/teamPreferences.coffee
```diff
@@ -33,7 +33,7 @@ exports.saveTeamPreferences = (teamPreferences, callback) ->
 
 exports.uploadTeamPhoto = (teamPreferencesId, file, callback) ->
```
dmccraw: Would it be better to rename `teamPreferencesId` to `teamPreferences` if it is an item?

dmccraw:
:+1: This looks good assuming that all the tests pass

src/collections/teamPreferences.coffee
matthopson: The reason for this is APIv3 actually takes a `team_preferences_id` explicitly (because it's a command). Eventually, I'd like to comb through all the SDK methods and allow them to accept both so it's always consistent, regardless of what v3 is expecting.

PR #64, opened 2015-10-01
Remove member statistic

matthopson:
- Should remove member statistics when deleting member.
- Takes care of a lingering issue (had it in another branch, but threw it away) where request is defined twice.


dmccraw:
:+1: Looks good

PR #66, opened 2015-10-13
Use new schema endpoint for loading collections

dmccraw:
@matthopson Take a look when you get back. I can walk you through the changes. We use the `/schemas` endpoint to load up the collections instead of getting each endpoint manually.


matthopson:
:+1: Excellent!

PR #69, opened 2015-10-26
File storage used persistence

matthopson:
When adding or removing team media, the file usage on the team object should be updated.


dmccraw:
:ship: it!

PR #71, opened 2015-10-30
if you load collections without items in plans, smsGateways, sports o…

dmccraw:
…r timeZones the sdk will crash


matthopson:
:+1:

PR #72, opened 2015-11-03
Bug fixes in setMediumAsMemberPhoto

matthopson:
- Should accept memberId (or member item).
- Should reload member item when done


dmccraw:
:ship: it!

PR #73, opened 2015-11-05
Reload member stats on bulk delete

matthopson:
* Small change in `persistence.coffee` to load all `memberStatistic` items when calling `bulkdDeleteSatisticData` to refresh all stat rankings and averages.
* Fix some lint warnings in other files.


amberwebb:
:+1:

PR #76, opened 2015-11-07
convert broadcastSms to broadcastAlert

dmccraw:
With the changes on the apiv3 to broadcast_alert, we need to change the sdk to reflect the changes.


semmons99:
:shipit:

matthopson:
@dmccraw Your base appears to be 2 versions behind.

dmccraw:
@matthopson I used pre-1.7.0 as the base. Is there a more recent branch?

matthopson:
That should be latest. But is already on 1.7.0-pre8.

matthopson:
@dmccraw Have you done a force push recently? History is all jacked up.

All these PRs are missing from `pre-1.7.0` 
https://github.com/teamsnap/teamsnap-javascript-sdk/pull/72
https://github.com/teamsnap/teamsnap-javascript-sdk/pull/73
https://github.com/teamsnap/teamsnap-javascript-sdk/pull/74

And your commit history looks a little funky for this PR as well.

screen shot 2015-11-07 at 5 58 51 pm

matthopson:
Also saying my local `pre-1.7.0` is now 12 commits ahead.

dmccraw:
Matt, I must have messed up the history. Can you force push your pre-1.7.0 version?

Thanks,
Dustin

matthopson:
Should be fixed up! :)

src/collections/broadcastAlerts.coffee
```diff
@@ -0,0 +1,36 @@
+exports.loadBroadcastAlerts = (params, callback) ->
+  if @isId params
+    params = id: params
```
matthopson: Should this be `teamId`?

dmccraw: broadcastAlerts, broadcastEmails and forumPosts all use an id as the default param for loading. The bulk loader passes a teamId so it will get picked up properly there.

dmccraw:
@matthopson good to go?

src/collections/broadcastAlerts.coffee
matthopson: I didn't realize this - other endpoints (that load more than one item of a type) use the teamId - since it essentially calls `collection.queryItems(params)` and searches on `teamId`

matthopson: Just coming back around to this - is it possible this is supposed to be `exports.loadBroadcastAlert` (singular) and take a `broadcastAlertId`?

It just seems like `loadBroadcastAlerts` should return all `broadcastAlert` items for the team.

dmccraw: `loadBroadcastAlerts`, `loadBroadcastEmails`, `loadBroadcastEmailAttachments`, `loadForumPosts`, `loadForumSubscriptions`, `loadForumTopics` all use id. Should we change all of these to use teamId? I am for making the change to use teamId for consistency but is this considered a breaking change for anyone using the sdk currently? I don't think we need the singular version unless we want to add singular versions to all the endpoints and then it's a lot of code for just duplicated code for something they can already use the load functions for already.

matthopson: My preference is to keep it consistent. 

I plan to go back through and add the singular versions in an SDK maintenance sweep (and right a few other little wrongs in the process).  Technically we've already broken it with the `broadcastSMS` change in APIv3 - and further with this PR.

dmccraw: All updated to use teamId.

matthopson:
:+1:

PR #78, opened 2015-11-11
Load member balance when new member is saved

amberwebb:
Fixes nj issue https://github.com/teamsnap/nextjenn-web/issues/1161


matthopson:
:+1: Dang, I totally thought we covered this.

PR #79, opened 2015-11-12
Load team fees when a roster member is created or deleted

amberwebb:
Fixes https://github.com/teamsnap/nextjenn-web/issues/1217


dmccraw:
:ship: it! Looks good.

matthopson:
Why reload `team_fees` when creating a new member?

amberwebb:
@matthopson When a new member is created or deleted, the team fee is not updated on the payments tab until refresh. Is there another way to do this?

matthopson:
The team fee should have nothing to do with the member, though.

amberwebb:
So the wrapper methods are specifically for the object. `saveMember` and `deleteMember` are only for objects related to member.

matthopson:
Yeah, and a `team_fee` object is completely independent of a member - if this is fixing an issue you're having, it's likely just masking something else. Let's look closer after DD.

amberwebb:
Ok. Closing.

matthopson:
:+1: I missed the `balance` portion of it. Disregard my dumb.

PR #81, opened 2015-11-17
Fix persistance problem when updating a stat that is used in another stat formula

nathanielvelez:
Added to persistence reloading of statisticData when saving a stat. Example:

In Game Stats: If StatA=StatB/2 and the user changes StatB to 20, StatA should show 10. This value wasn't being updated until the tab was reloaded.


dmccraw:
:ship: it
Does it make sense to make a bulk load call instead of 3 separate api calls?

matthopson:
@nathanielvelez Don't forget to `brunch build --production`

nathanielvelez:
Merged multi api calls into a single bulkLoad where possible, updated sdk version to pre13 and included production build.

lib/teamsnap.js
```diff
@@ -4159,20 +4159,26 @@ modifySDK = function(sdk) {
...
       teamId = templates[0].teamId;
-      sdk.loadMemberStatistics({
-        teamId: teamId
-      });
-      return sdk.loadStatisticAggregates({
-        teamId: teamId
-      });
+      return sdk.bulkLoad(teamId, "memberStatistic,statisticDatum,statisticAggregate");
+
+      /*
```
matthopson: Let's get these comments out of here.

lib/teamsnap.js
```diff
@@ -4159,20 +4159,26 @@ modifySDK = function(sdk) {
...
+    var teamId;
+    teamId = statisticDatum.teamId;
+    sdk.bulkLoad(teamId, "statisticDatum,statisticAggregate");
+    return sdk.loadMemberStatistics({
       statisticId: statisticDatum.statisticId
-    });
-    return sdk.loadStatisticAggregates({
-      teamId: statisticDatum.teamId
+
+      /*
```
matthopson: And these.

lib/teamsnap.js
```diff
@@ -4159,20 +4159,26 @@ modifySDK = function(sdk) {
...
+       */
     }
   });
   wrapSave(sdk, 'saveStatisticDatum', function(statisticDatum) {
-    sdk.loadMemberStatistics({
+    var teamId;
+    teamId = statisticDatum.teamId;
+    sdk.bulkLoad(teamId, "statisticDatum,statisticAggregate");
+    return sdk.loadMemberStatistics({
       statisticId: statisticDatum.statisticId
```
matthopson: We should actually be refreshing all `memberStatistic` items for the team, since ranking may change. Let's go ahead and fix that while we're in here.

matthopson: Nevermind - this is statistic-wide, so it shouldn't actually cause an issue. Disregard!

matthopson:
:+1:

PR #82, opened 2015-11-18
add reorder to teamMediaGroup

dmccraw:


src/collections/teamMediaGroups.coffee
```diff
@@ -22,3 +22,15 @@ exports.saveTeamMediaGroup = (teamMediaGroup, callback) ->
       "`teamMediaGroup.type` must be 'teamMediaGroup'"
 
   @saveItem teamMediaGroup, callback
+
+exports.reorderTeamMediaGroups = (teamId, teamMediaGroupIds, callback) ->
+  unless @isId teamId
+    throw new TSArgsError 'teamsnap.reorderTeamMediaGroups', '`teamId`
+      must be provided'
+  unless teamMediaGroupIds and typeof teamMediaGroupIds is 'object'
```
matthopson: We should probably use `Array.isArray teamMediaGroupIds` here since any object will validate here. Is this pulled from `reorderStatistics`? If so, I probably need to go back and fix that :grimacing:

matthopson:
Heh. I already wrote the SDK for this, but you beat me to PR. :stuck_out_tongue:

dmccraw:
@matthopson updated using `Array.isArray`

src/collections/statisticGroups.coffee
```diff
@@ -40,10 +40,12 @@ exports.reorderStatisticGroups = (teamId, statisticGroupIds, callback) ->
   unless @isId teamId
     throw new TSArgsError 'teamsnap.reorderStatisticGroups', '`teamId`
       must be provided'
-  unless statisticGroupIds and typeof statisticGroupIds is 'object'
-    throw new TSArgsError 'teamsnap.reorderStatisticGroups', 'You must provide
-    an array of ordered Statistic Group IDs'
+  unless statisticGroupIds and
+    typeof statisticGroupIds is 'object' and
```
matthopson: Not sure where my comment went:

No need to use `typeof` here. `Array.isArray()` will do this work for us.

dmccraw:
@matthopson updated

matthopson:
:+1:  Thanks for getting those others too

dmccraw:
No problem. I like having everything consistent. I will release both apiv3 and this in the morning.

PR #84, opened 2015-11-24
Stats new formula persistence

nathanielvelez:
Reloaded stat data after a new formula is created to be able to show the results of the formula if there is existing stat data.


src/persistence.coffee
```diff
@@ -334,6 +334,13 @@ modifySDK = (sdk) ->
         err
       ).callback callback
 
+  # Reload statisdic data when crating a new statistic with a formula
+  wrapSave sdk, 'saveStatistic', (statistic) ->
+    teamId = statistic.teamId
+    bulkLoadTypes = "statisticData,memberStatistic,
+                     teamStatistics,statisticAggregate"
```
matthopson: This _shouldn't_ be working for you as these aren't valid types. Have you tested?

Should be:
```coffeescript
`statisticDatum, memberStatistic, teamStatistic, statisticAggregate`
```

Also, please use single quotes on these to match other style. :wink:

matthopson:
:+1:

src/persistence.coffee
nathanielvelez: It was loading just the correct types which were the ones required for displaying the UI where I was testing. :fearful:

PR #86, opened 2015-12-01
Media facebook sharing

matthopson:
Adds support for `facebook_share_team_medium` and `facebook_share_team_media_group`


src/collections/teamMediaGroups.coffee
```diff
@@ -34,3 +35,38 @@ exports.reorderTeamMediaGroups = (teamId, teamMediaGroupIds, callback) ->
...
+    unless isSuppressedFromFeed? and typeof isSuppressedFromFeed is 'boolean'
+      throw new TSArgsError 'teamsnap.facebookShareMediaGroup', 'must include
+      boolean isSuppressedFromFeed'
+
+    params = {
+      teamMediaGroupId: teamMediaGroupId,
+      facebookPageId: facebookPageId,
+      albumName: albumName,
+      isSuppressedFromFeed: isSuppressedFromFeed
+    }
```
dmccraw: @matthopson As we make more and more values optional when do we need to start having more documentation on each of the functions?

dmccraw:
:ship: it! Looks good.

src/collections/teamMediaGroups.coffee
matthopson: I've got some running notes for a pretty big refactor on the SDK that will cover stuff like this. Might be a christmas project :D

PR #99, opened 2015-12-22
Ensure type is being properly converted for comparison

matthopson:
Since the addition of `type` on templates in APIv3, items being saved weren't diffing for changes properly as the `type` wasn't being properly converted to camelCase for the compare.


amberwebb:
:+1:

dmccraw:
@matthopson Good catch!

PR #103, opened 2016-01-06
Reload teamMediaGroups when saving / deleting teamMedia

matthopson:
Small persistence update to teamMedia for to reload teamMediaGroup on save / delete.


amberwebb:
CHIP it!

image

PR #114, opened 2016-02-08
Prevent manager from editing owner member item

matthopson:
Related to https://github.com/teamsnap/apiv3/pull/1786

A manager should not be able to edit an owner member item or its contacts. This fixes the `canEditItem` helper to validate this relationship on the `member` item.


amberwebb:
Looks good.

PR #116, opened 2016-02-15
Smart load

matthopson:
Refactors `bulkLoad` to accept new params for better scoping at load time.


docs/collections/teams.md
```diff
@@ -146,21 +146,44 @@ teamsnap.saveTeam(team).then(function(){
...
+// ~~~~~
+// Bulk load events and assignments related to eventId 1-6.
+bulkLoadParams = {
+  teamId: 1,
+  types: ['event', 'assignment'],
+  scopeTo: 'event',
+  event__id: '1,2,3,4,5,6'
+}
+teamsnap.bulkLoad(bulkLoadParams);
 ```
```
dmccraw: Love all the new documentation!!!!!

src/collections/teams.coffee
```diff
@@ -46,18 +46,30 @@ exports.deleteTeam = (team, callback) ->
 
 # Loads all items associated with a team, optionally limited by the types array
 exports.bulkLoad = (teamId, types, callback) ->
-  unless @isId teamId
-    throw new TSArgsError 'teamsnap.bulkLoad', 'teamId must be provided'
+  if typeof teamId is 'object'
+    loadParams = teamId
+    # Using params object (smartload)
+    unless @isId loadParams.teamId
```
dmccraw: Only checking for teamId? What happens if you don't send any types?

amberwebb: Doesn't seem to work unless you send types, no?

nathanielvelez: Types are check down bellow and will default to null, I assume if no types it will return no items, but it will have commands and the other info apiv3 gives you. I don't think it should throw an error if no types are given since the info it would provide is still useful while developing.

src/collections/teams.coffee
```diff
@@ -46,18 +46,30 @@ exports.deleteTeam = (team, callback) ->
...
-    types = null
+    if typeof types is 'function'
+      callback = types
+      types = null
 
   unless Array.isArray types
     types = @getTeamTypes()
     types.splice types.indexOf('availability'), 1
 
   params = teamId: teamId, types: types.map(@underscoreType).join(',')
```
dmccraw: What happens here if `types` is null?

src/collections/teams.coffee
matthopson: @dmccraw That's a good call. I was banking on it being undefined and being assigned in the code below if left off, but I didn't think about `Array.isArray()` evaluating an undefined var - it won't see it as `undefined` (which would eval to false, and be fine), it will throw a reference error.

Fixed the code where it will assign it (even if not present) so at the very least `type = undefined` will happen and `Array.isArray(undefined)` will actually return `false` and set types to `@getTeamTypes()`

matthopson:
@dmccraw Would you mind having another gander? Thanks!

dmccraw:
It looks good to me! :ship: it!

PR #117, opened 2016-02-24
Payment transaction

amberwebb:


src/collections/memberPayments.coffee
```diff
@@ -19,3 +19,17 @@ exports.saveMemberPayment = (memberPayment, callback) ->
     return @reject 'You must choose a member.', 'memberId', callback
 
   @saveItem memberPayment, callback
+
+exports.memberPaymentTransaction = (memberPaymentId, amount, callback) ->
+  unless @isItem memberPaymentId
```
matthopson: I hope I didn't steer you wrong here... 

we should drop the `@isItem` from this since we're validating that below.

amberwebb: So just `unless memberPaymentId` ?

matthopson:
Needs CHANGELOG and version bump.

src/collections/memberPayments.coffee
matthopson: Yup. Once we know there's a `memberPaymentId` we can move on to determining if they passed an `item` or `id` and do the appropriate thing from there.

src/collections/memberPayments.coffee
```diff
@@ -19,3 +19,17 @@ exports.saveMemberPayment = (memberPayment, callback) ->
...
+    throw new TSArgsError 'teamsnap.memberPaymentTransaction', "must provide
+     a `memberPaymentId`"
+  if @isItem memberPayment
+    memberPayment = memberPayment.id
+  unless amount
```
nathanielvelez: Should't we test if the it throw an error if the object is not a memberPayment?

matthopson: Nah, the command takes a `member_payment_id` but in the SDK we're going to allow a `memberPayment` object - since that's our general convention. If someone does pass in an object instead, we're just going to strip the `id` off it.

amberwebb:
@matthopson Please re-review.

lib/teamsnap.js
```diff
@@ -5805,7 +5823,7 @@ ref = require('./model'), Collection = ref.Collection, Item = ref.Item;
 require('./errors');
 
 TeamSnap = (function() {
-  TeamSnap.prototype.version = '1.10.0';
+  TeamSnap.prototype.version = '1.10.1';
```
matthopson: Need to bump minor version instead of patch since we changed the API. -  `1.11.0`

matthopson:
Sorry, I thought of one other thing... we probably want to go ahead and replicate what we're doing in persistence with `saveMemberPayment`, reloading `memberBalance` and `teamFee`.

https://github.com/teamsnap/teamsnap-javascript-sdk/blob/master/src/persistence.coffee#L493

Since it's possible the old method is still being used to do the same thing, we should leave it in place in persistence.

src/collections/memberPayments.coffee
```diff
@@ -19,3 +19,17 @@ exports.saveMemberPayment = (memberPayment, callback) ->
     return @reject 'You must choose a member.', 'memberId', callback
 
   @saveItem memberPayment, callback
+
+exports.memberPaymentTransaction = (memberPaymentId, amount, callback) ->
+  unless memberPaymentId
+    throw new TSArgsError 'teamsnap.memberPaymentTransaction', "must provide
+     a `memberPaymentId`"
+  if @isItem memberPayment
+    memberPayment = memberPayment.id
```
matthopson: We're using `memberPaymentId` as the var. Should be replaced on 27 and 28

amberwebb:
@matthopson Regarding this comment, https://github.com/teamsnap/teamsnap-javascript-sdk/pull/117#issuecomment-188816858. Why do you think it's necessary? In njw, it's not changing the memberPayment object until saveMemberPayment is called, which is loading `teamFee` and `memberBalance`

matthopson:
@amberwebb This command updates the amount owed and the amount paid, which is stored on the `memberPayment` object... this also affects the `teamFee.balance` amount, so it's imperative these are reloaded to reflect correct balances.

If we're just adding a payment transaction, then we shouldn't even be calling `saveMemberPayment`

docs/collections/member_payments.md
```diff
@@ -47,3 +52,22 @@ Saves a `memberPayment` item.
...
+
+---
+
+
+<a id="memberPaymentTransaction"></a>
+## `memberPaymentTransaction(memberPaymentId, amount, callback)`
+Creates a `memberPaymentTransaction`.
+
+### Params
+* `memberPaymentId`: [id, object] - `memberPaymentId` or `memberPayment` item to be saved.
```
matthopson: Need to add `amount` and `note` params

src/persistence.coffee
```diff
@@ -682,6 +684,18 @@ modifySDK = (sdk) ->
           sdk.loadContactEmailAddresses({contactId: contactId}).then -> result
     ).callback callback
 
+  # Update memberBalance and teamFee when calling memberPaymentTransaction
+  wrapMethod sdk, 'memberPaymentTransaction', (memberPaymentTransaction) ->
+    (memberPaymentId, callback) ->
```
matthopson: Missing `amount` and `note` params.

src/persistence.coffee
```diff
@@ -682,6 +684,18 @@ modifySDK = (sdk) ->
...
+      memberPaymentTransaction.call(this, memberPaymentId).then((result) ->
```
matthopson: Same as above

src/teamsnap.coffee
```diff
@@ -3,7 +3,7 @@ promises = require './promises'
 require './errors'
 
 class TeamSnap
-  version: '1.10.0'
+  version: '1.11.0'
```
matthopson: Don't forget to also bump version in `package.json`

test/memberPayments.coffee
```diff
@@ -1,7 +1,29 @@
...
+    }
+    done()
+
+  describe 'loadMemberPayments', ->
+    it 'should be able to load all member payments for team', ->
+      teamsnap.loadMemberPayments team.id, (err, result) ->
+        expect(err).to.be.null
+        result.should.be.an('array')
+
+  describe 'memberPaymentTransaction', ->
```
matthopson: Should we test for note as well?

src/persistence.coffee
amberwebb: I didn't think I needed to add the note and amount params since I only really need the memberPaymentId to find the memberBalance and teamFee to reload. Is this really necessary or more of a consistency thing?

src/collections/memberPayments.coffee
```diff
@@ -19,3 +19,17 @@ exports.saveMemberPayment = (memberPayment, callback) ->
     return @reject 'You must choose a member.', 'memberId', callback
 
   @saveItem memberPayment, callback
+
+exports.memberPaymentTransaction = (memberPaymentId, amount, note, callback) ->
+  unless memberPaymentId
```
matthopson: I don't recall if `note` is optional (I believe it is)... so we should test if the `note` is a function and set that to our `callback`

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### Feb 24, 2016 // Version 1.11.0
+- Adds `memberPaymentTransaction` method
+- Adds wrapper to persistence for loading teamFee and memberBalance in` memberPaymentTransaction`
+---
```
matthopson: Missing line break here (but it looks like the prev entry did the same).

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### Feb 24, 2016 // Version 1.11.0
+- Adds `memberPaymentTransaction` method
+- Adds wrapper to persistence for loading teamFee and memberBalance in` memberPaymentTransaction`
```
matthopson: wrap `teamFee` and `memberBalance` in backtics, and fix spacing in `memberPaymentTransaction`

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### Feb 24, 2016 // Version 1.11.0
```
matthopson: Today is the 26th

docs/collections/member_payments.md
```diff
@@ -47,3 +52,22 @@ Saves a `memberPayment` item.
 // Saves memberPayment item.
 teamsnap.saveMemberPayment(memberPayment);
 ```
+
+
+---
+
+
+<a id="memberPaymentTransaction"></a>
+## `memberPaymentTransaction(memberPaymentId, amount, callback)`
```
matthopson: Add Note `param`

docs/collections/member_payments.md
```diff
@@ -47,3 +52,22 @@ Saves a `memberPayment` item.
...
+
+### Params
+* `memberPaymentId`: [id, object] - `memberPaymentId` or `memberPayment` item to be saved.
+* `callback`: [function] - callback to be executed when the operation completes.
+
+### Examples
+```javascript
+// ~~~~~
+// Creates memberPaymentTransaction.
+teamsnap.memberPaymentTransaction(1, 5.00);
```
matthopson: Example should include a note.

matthopson:
@amberwebb Couple of tips for SDK:
1) I typically don't build and commit `.js` files until approved - this just makes reviewing it easier as well as making changes since you don't have to keep building that now that it's part of the commit history.
2) I like to try and squash my commits down so the commit log is easily readable - this is mostly the pattern I follow:

- Add some feature (.coffee)
- Update persistence (optional - this could be in the above)
- Add tests for feature (optional - could also be in the above)
- Update CHANGELOG and bump version (`teamsnap.coffee` and `package.json`)
APPROVAL
- Build v1.x.x
MERGE

test/memberPayments.coffee
amberwebb: Yes, but testing needs work. I don't think this needs to be a part of this PR.

src/persistence.coffee
matthopson: This literally wraps our `memberPaymentTransaction` method... without passing the into the wrapper, they won't be available when you actually `.call` the method.

nathanielvelez:
@matthopson That pattern you suggested here could be nice addition to the repo wiki. Will save it in my notes.

PR #118, opened 2016-02-26
memberPaymentTransaction, persistence, tests, and docs

amberwebb:
Version changes in teamsnap.coffee and package.json in next PR with teamsnap.js


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### Feb 26, 2016 // Version 1.11.0
+- Adds `memberPaymentTransaction` method.
+- Adds persistence wrapper to `memberPaymentTransaction` to reload `teamFee` and `memberBalance`.
+---
```
matthopson: Add line break here between the last log item and the dashes.

docs/collections/member_payments.md
```diff
@@ -47,3 +48,24 @@ Saves a `memberPayment` item.
...
+---
+
+
+<a id="memberPaymentTransaction"></a>
+## `memberPaymentTransaction(memberPaymentId, amount, note, callback)`
+Creates a `memberPaymentTransaction`.
+
+### Params
+* `memberPaymentId`: [id, object] - `memberPaymentId` or `memberPayment` item to be saved.
+* `amount`: [int] - `memberPayment.amountPaid`.
```
matthopson: Maybe just describe the amount ("The transaction amount" or something) - we don't really need to reference the `memberPayment` item here as it just makes it confusing - and technically it takes a `float`, right?

docs/collections/member_payments.md
```diff
@@ -47,3 +48,24 @@ Saves a `memberPayment` item.
...
+* `note`: [string] - optional string.
```
matthopson: For params, I've been appending '(optional)' to the end of the description. So maybe something like:

`A note for the transaction (optional)`

docs/collections/member_payments.md
amberwebb: ` amount`: [int, float] - The transaction amount.` Yes, it does take a float or int.

test/memberPayments.coffee
```diff
@@ -1,7 +1,27 @@
 describe 'Member Payments', ->
```
matthopson: Does this test work on a fresh team with no `teamFee` or  `memberPayment` objects? I suspect these will fail since it won't already have existing objects to run against.

amberwebb: Yes, they will fail if testing against a team without those objects. The problem I have with testing is that I am testing against what is in my local. Testing as whole in the SDK needs some work. It's not important to fix partial payments, so I plan to come back to it.

amberwebb:

amberwebb:
Commits successfully squashed. :)

matthopson:
Missing version bump in `package.json` and `teamsnap.coffee` (1.11.0)

matthopson:
:+1: Build and merge!

PR #119, opened 2016-03-02
Add loadPaymentNotes to saveMemberPayment persistence wrapper

amberwebb:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### March 2, 2016 // Version 1.12.0
```
matthopson: Should be a patch

src/persistence.coffee
```diff
@@ -498,9 +498,11 @@ modifySDK = (sdk) ->
       saveMemberPayment.call(this, memberPayment).then((result) ->
         memberId = result.memberId
         teamFeeId = result.teamFeeId
+        teamId = result.teamId
         promises.when(
           sdk.loadMemberBalances(memberId: memberId)
           sdk.loadTeamFees(id: teamFeeId)
+          sdk.loadPaymentNotes(teamId: teamId)
```
matthopson: We should probably not load the whole team's notes - at the very least we could do it on member, but we should be able to use the teamFee as well and be more pacific.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
...
+- Reloads `loadPaymentNotes` in `saveMemberPayment` method in persistence wrapper.
```
matthopson: "Reloads `paymentNotes`..."

src/persistence.coffee
```diff
@@ -501,6 +501,7 @@ modifySDK = (sdk) ->
         promises.when(
           sdk.loadMemberBalances(memberId: memberId)
           sdk.loadTeamFees(id: teamFeeId)
+          sdk.loadPaymentNotes(teamFeeId: teamFeeId)
```
matthopson: Was looking at `v3` - it seems we can query on `memberPaymentId`
https://github.com/teamsnap/apiv3/blob/master/app/payment_notes/payment_note_serializer.rb#L34

amberwebb:

matthopson:
:+1:

PR #120, opened 2016-03-02
Add error message if amount is null, update patch version and docs

amberwebb:


src/collections/memberPayments.coffee
```diff
@@ -27,8 +27,7 @@ exports.memberPaymentTransaction = (memberPaymentId, amount, note, callback) ->
   if @isItem memberPaymentId
     memberPaymentId = memberPaymentId.id
   unless amount
-    throw new TSArgsError 'teamsnap.memberPaymentTransaction', "must provide
-     an `amount`"
+    return @reject 'You must add an amount.'
```
nathanielvelez: Normally we also return the field and the callback when `@reject()`
`return @ reject 'You must add an amount.', 'amount', callback`

amberwebb:
@nathanielvelez Please review again.

nathanielvelez:
:+1:

PR #121, opened 2016-03-07
Import from team

amberwebb:


nathanielvelez:
:+1:

src/collections/members.coffee
```diff
@@ -145,3 +145,27 @@ exports.canEditItem = (member, team, item) ->
     item.href is member.href
   else
     item.links.member.href is member.href
+
+exports.importMembersFromTeam = (memberIds, teamId, callback) ->
```
matthopson: Curious, what does this method return when successful?

docs/collections/members.md
```diff
@@ -292,3 +294,40 @@ if(teamsnap.canEditItem(member, team, availability)){
...
+---
+
+
+<a id="loadImportableMembers"></a>
+## `loadImportableMembers(userId, includeArchivedTeams)`
+Creates a list of importable members by current member's userId.
+
+## Params
+* `userId`: [id] - `userId` of current member.
+* `includeArchivedTeams`: [bool] - include or exclude members from archived teams.
```
matthopson: optional?

src/collections/members.coffee
amberwebb: This imports the members into the current team.

matthopson: Right, but what I'm wondering is what's in the response body... is it a list of member that are bing imported? Or the team? Or what?

src/collections/members.coffee
```diff
@@ -145,3 +145,27 @@ exports.canEditItem = (member, team, item) ->
...
+    `memberIds`'
+  unless @isItem teamId
+    throw new TSArgsError 'teamsnap.importMembersFromTeam', "must provide
+     a teamId"
+  if @isItem teamId
+    teamId = teamId.id
+  params = sourceMemberIds: memberIds, destinationTeamId: teamId
+
+  @collections.members.exec('importFromTeam', params)
+    .pop().callback callback
```
matthopson: So if we're returning an array of members that were added, do we want to pop() that list? Or return the whole array of members? I feel like we want to return all the members.

amberwebb: Why do you think so?

matthopson: Doesn't it make sense to get the full return vs a piece of it? We may not do this in our app, but if someone working with v3 was expecting to see the entire list of imported members in the response, but didn't with the SDK, we'd be tricking them up. You may, for instance, want to take the response and display a confirmation: "The following members were imported..." or something.

amberwebb:

matthopson:
Looks good :+1:

PR #122, opened 2016-03-17
Add method to sort forumTopics by last post

amberwebb:


nathanielvelez:
:+1:

src/collections/forumTopics.coffee
```diff
@@ -32,3 +32,14 @@ exports.deleteForumTopic = (forumTopic, callback) ->
       be provided'
 
   @deleteItem forumTopic, callback
+
+exports.getForumTopicPostsSort = ->
+  (itemA, itemB) ->
+    valueA = new Date(itemA.forumPosts.last().createdAt)
```
matthopson: How does this hold up when there are no posts for a given topic?

src/collections/forumTopics.coffee
```diff
@@ -32,3 +32,14 @@ exports.deleteForumTopic = (forumTopic, callback) ->
       be provided'
 
   @deleteItem forumTopic, callback
+
+exports.getForumTopicPostsSort = ->
```
matthopson: I guess we don't care about being able to reverse the sort order?

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### March 17, 2016 // Version 1.13.0
+- Adds `getForumTopicPostsSort` method.
```
matthopson: This method only sorts `forumTopic` items, not `posts`.

src/collections/forumTopics.coffee
amberwebb: There's always a forumPost. `forumTopic` is a container for `forumPost`. So when you create a topic, you are also creating a post.

src/collections/forumTopics.coffee
amberwebb: No, we don't. ;)

In nextjenn, if I specify `desc` in the sort, it reverses the order.

src/collections/forumTopics.coffee
matthopson: But it isn't necessarily loaded (for instance in the case of Onion).

amberwebb: Do you mean forumPosts? Not following why forumPosts wouldn't be loaded if forumTopics are.

PR #128, opened 2016-05-03
Reload loadMessages in saveBroadcastAlert

amberwebb:


src/persistence.coffee
```diff
@@ -715,6 +716,12 @@ modifySDK = (sdk) ->
         ).then -> result
       ).callback callback
 
+  wrapMethod sdk, 'saveBroadcastAlert', (saveBroadcastAlert) ->
+    (broadcastAlert, callback) ->
+      saveBroadcastAlert.call(this, broadcastAlert).then((result) ->
+        sdk.loadMessages({messageId: result.id})
```
matthopson: We have a `message.id` and a `message.messageId`? And they aren't the same thing? AND `message.messageId` === `broadcastAlert.id`?

CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### May 2, 2016 // Version 1.13.3
+- Reloads `loadMessages` in `saveBroadcastAlert` method in persistence wrapper.
```
matthopson: Tiny nit: this reads kinda funny... could you say something like "Reloads `messages` for `broadcastAlert` in `saveBroadcastAlert`...."

src/persistence.coffee
amberwebb: Well when you put it that way it sounds confusing. :)

`message.id` and `message.messageId` are not the same. `messageId` is `broadcastAlert.id`.

amberwebb: But I'm not sure why we even have a `loadMessages`. Can't we just use `loadBroadcastAlerts` ? @randy-girard Thoughts on this?

glanotte: I am going to have to concur, that is a bit confusing.

amberwebb: I see why we have `loadMessages`. Are we planning on renaming this property?

randy-girard: Messages are the main object now. They are the instance of the broadcast for a member/contact. Right now they only point broadcast alerts, but they will eventually also point to broadcast emails. The reason I added messageId to a message, is to point back to the actual content of the message (ie broadcastAlert/broadcastEmail/forumPost possibly). We can chat if this doesn't make sense.

randy-girard:
I think you also need to reload messageData as well, since if you send a message to yourself, you will now have an unread message in your inbox, but if we don't reload it, the member won't see the visual badge on the alert tab.

amberwebb:

src/persistence.coffee
```diff
@@ -715,6 +716,13 @@ modifySDK = (sdk) ->
         ).then -> result
       ).callback callback
 
+  wrapMethod sdk, 'saveBroadcastAlert', (saveBroadcastAlert) ->
+    (broadcastAlert, callback) ->
+      saveBroadcastAlert.call(this, broadcastAlert).then((result) ->
+        sdk.loadMessageData({memberId: result.member.id})
```
randy-girard: Its possible the person sending the broadcast could be a contact and not a member, in which case we would want to update their message data.

matthopson: Will it only return one or the other?
Could you do something like:
```coffeescript
personId = result.member?.id || result.contact.id
```

matthopson: Oh but you need to account for `memberId`, duh

matthopson:
```coffeescript
if result.member?
  params = {memberId: result.member.id}
else
  params = {contactId: result.contact.id}

sdk.loadMessageData(params)
```

Not sure if that's going to work (without seeing the response), but if it's what I think, something along those lines should work.

matthopson: Sorry, one more thing... ☝️ that shouldn't actually work.

Should be using `result.memebrId` / `result.contactId` since `member` and `contact` are linked to the object - not actually part of the object's template.

src/persistence.coffee
```diff
@@ -719,7 +719,8 @@ modifySDK = (sdk) ->
   wrapMethod sdk, 'saveBroadcastAlert', (saveBroadcastAlert) ->
     (broadcastAlert, callback) ->
       saveBroadcastAlert.call(this, broadcastAlert).then((result) ->
-        sdk.loadMessages({messageId: result.id})
+        sdk.loadMessageData({memberId: result.member.id})
```
amberwebb: Yeah, I got that backwards. Thanks for catching that.

src/persistence.coffee
amberwebb: I actually got that backwards. `messageData` has a member object.

amberwebb: Looks like we are not loading it in the collection. 

https://github.com/teamsnap/apiv3/blob/master/app/broadcast_alerts/broadcast_alert_serializer.rb#L15

amberwebb:
@randy-girard @matthopson Please review again.

matthopson:
👍 Looks good to me

src/persistence.coffee
```diff
@@ -715,6 +716,17 @@ modifySDK = (sdk) ->
         ).then -> result
       ).callback callback
 
+  wrapMethod sdk, 'saveBroadcastAlert', (saveBroadcastAlert) ->
+    (broadcastAlert, callback) ->
+      saveBroadcastAlert.call(this, broadcastAlert).then((result) ->
+        if result.member?
```
randy-girard: Might want to also check for `result.divisionMember?` here? You can still use `memberId: result.id`.

matthopson: Didn't even think about divisionMember. :/

randy-girard:
👍  Looks good

PR #129, opened 2016-05-23
Reload members in importMembersFromTeam

amberwebb:
We have an issue where member information is not displaying until after refresh of the page once `importMemberFromTeam` has been called. This is reloading the specified members to display the information in the UI.

I think we could only reload the member object, not needing to reload email addresses, phone numbers or contacts in the member explicitly.

### DIFF
--- a/src/persistence.coffee
+++ b/src/persistence.coffee
@@ -195,7 +195,9 @@ modifySDK = (sdk) ->
   # 10. deleteForumTopic needs to delete all related posts
   # 11. memberEmailAddresses need to reload when invite is sent
   # 12. teamFee and memberBalance needs to reload after transaction
-  # 13. members need to be reloaded after importMembersFromTeam
+  # 13. members, contacts, memberEmailAddresses, memberPhoneNumbers,
+  #contactEmailAddresses, contactPhoneNumbers need to be reloaded 
+  #after importMembersFromTeam
   # 14. messages and messageData need to reload after saveBroadcastAlert
 
   # Load related records when a member is created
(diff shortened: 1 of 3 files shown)


src/persistence.coffee
```diff
@@ -708,11 +707,7 @@ modifySDK = (sdk) ->
       .then((result) ->
         memberIds = result.map (member) -> member.id
         promises.when(
-          sdk.loadMemberEmailAddresses({memberId: memberIds})
-          sdk.loadContactEmailAddresses({memberId: memberIds})
-          sdk.loadContacts({memberId: memberIds})
-          sdk.loadMemberPhoneNumbers({memberId: memberIds})
-          sdk.loadContactPhoneNumbers({memberId: memberIds})
+          sdk.loadMembers({id: memberIds})
```
nathanielvelez: Did you test that all the emails and the phone numbers are available without reloading them here? Not sure if just loading members will grab all the related types.

amberwebb: Thanks @nathanielvelez. You're right, these need to be loaded. Should I also reload the member here?

amberwebb: And also, should contacts be loaded before `contactEmailAddresses` and `contactPhoneNumbers` ?

nathanielvelez: I would assume that yes, If the members were added thru the import endpoint, they probably would't be available without reloading. About if the order on how things are loaded affect the linking I will assume they shouldn't, but I am not sure. Maybe Mr. @matthopson could clarify that.

amberwebb: We've had this conversation before and it was determined that the order of loading shouldn't affect things, but it does. If contacts are loaded first, the `contactMemberEmailAddress` and `contactPhoneNumber` loads.

nathanielvelez:
👍

PR #130, opened 2016-06-01
Revert "Reload members in importMembersFromTeam"

amberwebb:
Reverts teamsnap/teamsnap-javascript-sdk#129


nathanielvelez:
👍

knappe:
@amberwebb Would you please update this PR to include why this was reverted?  It will make the history a lot easier to follow if we know why this was reverted.

PR #132, opened 2016-06-01
Reload members after importing from another team

amberwebb:


nathanielvelez:
It still doesn't have teamsnap.js

nathanielvelez:
👍

PR #136, opened 2016-06-20
Team photos

matthopson:
Adds `teamPhotos` and `memberPhotos` methods for v3 endpoints.


docs/collections/team_photos.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Examples
+```javascript
+// ~~~~~
+// Loads a teamPhoto for `id: 1`.
+teamsnap.loadTeamPhoto(1);
+
+// ~~~~~
+// Loads a teamPhoto for `teamId: 1` and specifies the width to return.
+teamsnap.loadTeamPhoto({teamId: 1, width: 200});
```
nathanielvelez: Can a team have more than 1 photo. In that case, would this method return all the photos for the team. the first one or a random one? How is this different from `loadTeamPhotos`?

test/memberPhotos.coffee
```diff
@@ -0,0 +1,7 @@
+describe 'Member Photo', ->
+
+  it 'should be able to load all members photos for a team', (done) ->
+    teamsnap.loadMemberPhotos team.id, (err, result) ->
+      expect(err).to.be.null
+      result.should.be.an('array')
+      done()
```
nathanielvelez: Shouldn't we also test the `loadMemberPhoto` method?

test/teamPhotos.coffee
```diff
@@ -0,0 +1,7 @@
+describe 'Team Photo', ->
+
+  it 'should be able to load all team photos for a team', (done) ->
+    teamsnap.loadTeamPhotos team.id, (err, result) ->
+      expect(err).to.be.null
+      result.should.be.an('array')
+      done()
```
nathanielvelez: Shouldn't we also test the `loadTeamPhoto` method?

docs/collections/team_photos.md
matthopson: My guess is we won't ever allow more than one team photo, so it may not be needed - but I'll need to confirm.

test/memberPhotos.coffee
matthopson: I'm just still not sure how to test these photo uploads. Loading the singular version will return `undefined` if it doesn't exist, I believe, which makes it even harder to test if you can't test an upload first. Will need to double-check...

test/teamPhotos.coffee
matthopson: Same as ☝️

matthopson:
@nathanielvelez I've beefed up the tests a little. Please re-review.

docs/collections/team_photos.md
matthopson: After some thought, it occurred to me, we will expect to see multiple `teamPhoto` items if querying multiple teams - so this needs to stay.

nathanielvelez:
👍

test/memberPhotos.coffee
```diff
@@ -1,7 +1,31 @@
...
+
 
   it 'should be able to load all members photos for a team', (done) ->
     teamsnap.loadMemberPhotos team.id, (err, result) ->
       expect(err).to.be.null
       result.should.be.an('array')
+      result[0].should.have.property('type', 'memberPhoto')
+      done()
+
+  it 'should be able to load a single team photo', (done) ->
```
nathanielvelez: This should be called `ld be able to load a single member photo`

nathanielvelez:
@matthopson A small error on the description of the loadMemberPhoto test. Everything else looks good.

PR #137, opened 2016-06-23
Assignments mvp changes

amberwebb:
This adds methods to volunteer for, opt out of, and send assignments emails.

### DIFF
--- a/docs/collections/assignments.md
+++ b/docs/collections/assignments.md
@@ -146,22 +146,26 @@ assignmentArray.sort(teamsnap.getAssignmentSort());
 
 <a id="optOutOfAssignments"></a>
 ## `optOutOfAssignments(assignmentIds, callback)`
-Unsets a member from `assignment`.
+Unsets the current active member from `assignment`.
 
 ### Params
-* `assignmentIds`: [ids] - assignment ids to unset a member from.
+* `assignmentIds`: [string, object] - string of assignmentIds to unset the
+   current active member from. can also be an object.
 * `callback`: [function] - callback to be executed when the operation completes.
 
 ### Examples
 ```javascript
 // ~~~~~
-//Unsets a member from an `assignment`.
-teamsnap.optOutOfAssignments([9, 10]);
+//Unsets the current active member from an `assignment` based on `assignmentId` .
+teamsnap.optOutOfAssignments("1");
+
+// ~~~~~
+//Unsets the current active member from multiple `assignments` based on `assignmentIds` .
+teamsnap.optOutOfAssignments("1, 2, 3");
 ```
 
 
 ---
-
 
 
 <a id="volunteerForAssignments"></a>
@@ -169,15 +173,20 @@ teamsnap.optOutOfAssignments([9, 10]);
 Sets a member to an assignment with a null memberId.
 
 ### Params
-* `assignmentIds`: [ids] - assignment ids to set a member.
-* `memberId`: [id] - id of member to be assigned.
+* `assignmentIds`: [array, object] - list of assignment ids to set the current
+   active member to. Can also be an assignment object.
+* `memberId`: [int, object] - memberId or member object of current active member.
 * `callback`: [function] - callback to be executed when the operation completes.
 
 ### Examples
 ```javascript
 // ~~~~~
-//Sets a member to an assignment.
+//Sets the current active member to multiple assignments.
 teamsnap.volunteerForAssignments([4, 5], 1);
+
+// ~~~~~
+//Sets the current active member to one assignment.
+teamsnap.volunteerForAssignments([2], 1);
 ```
 
 
@@ -189,8 +198,9 @@ teamsnap.volunteerForAssignments([4, 5], 1);
 Sends assignments emails for events to members on a team.
 
 ### Params
-* `teamId`: [id] - team id.
-* `eventIds`: [ids] - array of ids for selected events.
+* `teamId`: [int] - teamId of current team.
+* `eventIds`: [array, object] - list of eventIds for selected events. Can also
+   be an event object.
 * `message`: [string] - message for email. (optional)
 * `callback`: [function] - callback to be executed when the operation completes.
 
--- a/docs/collections/assignments.md
+++ b/docs/collections/assignments.md
@@ -7,6 +7,9 @@
 - [saveAssignment](#saveAssignment)
 - [deleteAssignment](#deleteAssignment)
 - [getAssignmentSort](#getAssignmentSort)
+- [optOutOfAssignments](#optOutOfAssignments)
+- [volunteerForAssignments](#volunteerForAssignments)
+- [sendAssignmentEmails](#sendAssignmentEmails)
 
 
 ---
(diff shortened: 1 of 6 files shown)


src/collections/assignments.coffee
```diff
@@ -59,4 +53,36 @@ exports.getAssignmentSort = (reverse) ->
       else if valueA and !valueB then -1
       else if valueA > valueB then 1
       else if valueA < valueB then -1
-      else 0
\ No newline at end of file
+      else 0
+
+exports.optOutOfAssignments = (assignmentIds, callback) ->
```
matthopson: I feel like I'm missing a piece of the puzzle. Why do we need this method?

amberwebb: A manager assigns a task to a user. The user can unassign themselves.

matthopson: Ah, so it isn't actually deleting it. Just nulling the member reference?

Is it extrapolating the `memberId` from the request / `assignmentId`?

amberwebb: Exactly.

matthopson: Cool, last question(s).

- Does it accept a single ID in addition to the list of IDs (basing this on usage in NJW).
- We have some built-in ID validation (`@isId`) if you want to use it.
- Can we allow it to accept an `assignment` object to be a little more consistent with other methods?

*Random Note: I'm a little bummed we went with plural terminology in v3 - again, I think for consistency (and maybe a little more intuitive).

src/collections/assignments.coffee
```diff
@@ -59,4 +53,36 @@ exports.getAssignmentSort = (reverse) ->
...
+  unless assignmentIds
+    throw new TSArgsError 'teamsnap.optOutOfAssignments', 'must include
+    `assignmentIds`'
+  params = assignmentIds: assignmentIds
+  @collections.assignments.exec('optOutOfAssignments', params)
+  .pop().callback callback
+
+exports.volunteerForAssignments = (assignmentIds, memberId, callback) ->
```
matthopson: It seems a little strange for me that you'd be required to use a `memberId` to volunteer - but not to un-volunteer.

matthopson: Also same notes as above kinda apply here. Single id, vs multiple, allow objects, etc...

src/collections/assignments.coffee
amberwebb: It accepts both a single id and a list of ids.

src/collections/assignments.coffee
```diff
@@ -59,4 +53,36 @@ exports.getAssignmentSort = (reverse) ->
...
+  unless @isItem memberId
+    throw new TSArgsError 'teamsnap.volunteerForAssignments', "must provide
+     a memberId"
+  if @isItem memberId
+    memberId = memberId.id
+
+  params = assignmentIds: assignmentIds, memberId: memberId
+  @collections.assignments.exec 'volunteerForAssignments', params, callback
+
+exports.sendAssignmentEmails = (teamId, eventIds, message, callback) ->
```
matthopson: Do we not require `message`?

src/collections/assignments.coffee
amberwebb: Not sure I'm following why you think that is strange. When un-volunteering, the memberId is pulled from the assignmentId. Can you explain a little more?

src/persistence.coffee
```diff
@@ -738,6 +740,12 @@ modifySDK = (sdk) ->
           params = {contactId: result.contactId}
         sdk.loadMessageData(params)
         sdk.loadMessages({messageSourceId: result.id})
+
+  wrapMethod sdk, 'optOutOfAssignments', (optOutOfAssignments) ->
```
matthopson: You shouldn't have to do this. The response object should have the updated `assignment` with a `null` `memberId`, which _should_ unlink the member item.

docs/collections/assignments.md
```diff
@@ -136,3 +139,64 @@ Sorts an array of assignments by member.
...
 assignmentArray.sort(teamsnap.getAssignmentSort());
 ```
+
+
+---
+
+
+<a id="optOutOfAssignments"></a>
+## `optOutOfAssignments(assignmentIds, callback)`
+Unsets a member from `assignment`.
```
matthopson: Should be more descriptive to note it's unsetting the current active `member` (you) from the assignment.

src/collections/assignments.coffee
amberwebb: No. It is not required.

src/persistence.coffee
amberwebb: It does have that in the response, but it's not unlinking the item. I was really not sure about doing this here. Just reloading the assignment was not clearing the member item.

docs/collections/assignments.md
```diff
@@ -136,3 +139,64 @@ Sorts an array of assignments by member.
...
+
+### Params
+* `assignmentIds`: [ids] - assignment ids to unset a member from.
```
matthopson: - `[_type_]` should be string, object, array, function, integer, etc...
- Should note an assignmentId, list of assignmentIds, or object (if applicable)

docs/collections/assignments.md
```diff
@@ -136,3 +139,64 @@ Sorts an array of assignments by member.
...
+* `callback`: [function] - callback to be executed when the operation completes.
+
+### Examples
+```javascript
+// ~~~~~
+//Unsets a member from an `assignment`.
```
matthopson: "Unsets current active member..."

docs/collections/assignments.md
amberwebb: We do that in a few other places `[ids]`. Probably because I wrote the docs. :)

docs/collections/assignments.md
```diff
@@ -136,3 +139,64 @@ Sorts an array of assignments by member.
...
+teamsnap.optOutOfAssignments([9, 10]);
```
matthopson: I was under the impression (based on usage in NJW) that this accepts a list (string) of ids, not an array.

docs/collections/assignments.md
```diff
@@ -136,3 +139,64 @@ Sorts an array of assignments by member.
...
+---
+
+
+
+<a id="volunteerForAssignments"></a>
+## `volunteerForAssignments(assignmentIds, memberId, callback)`
+Sets a member to an assignment with a null memberId.
+
+### Params
+* `assignmentIds`: [ids] - assignment ids to set a member.
```
matthopson: - `[_type_]`
- Should note an assignmentId, list (or array) of assignmentIds, or object (if applicable)

docs/collections/assignments.md
```diff
@@ -136,3 +139,64 @@ Sorts an array of assignments by member.
...
+* `memberId`: [id] - id of member to be assigned.
```
matthopson: - `[_type_]`
- Should note a memberId or member object

docs/collections/assignments.md
```diff
@@ -136,3 +139,64 @@ Sorts an array of assignments by member.
...
+
+---
+
+
+<a id="sendAssignmentEmails"></a>
+## `sendAssignmentEmails(teamId, eventIds, message, callback)`
+Sends assignments emails for events to members on a team.
+
+### Params
+* `teamId`: [id] - team id.
```
matthopson: - [_type_]
- `teamId`

docs/collections/assignments.md
```diff
@@ -136,3 +139,64 @@ Sorts an array of assignments by member.
...
+* `eventIds`: [ids] - array of ids for selected events.
```
matthopson: - [_type_]

docs/collections/assignments.md
amberwebb: You're right. And I talked about this yesterday with JJ. Originally it took an array but now accepts a string of ids. I'll fix that.

I've been sitting on this code for a while and should have remembered to change that.

src/collections/assignments.coffee
matthopson: I get it. I was under the impression the request could be used to determine the member - but I guess we aren't doing that here. So it makes sense.

docs/collections/assignments.md
```diff
@@ -146,38 +146,47 @@ assignmentArray.sort(teamsnap.getAssignmentSort());
 
 <a id="optOutOfAssignments"></a>
 ## `optOutOfAssignments(assignmentIds, callback)`
-Unsets a member from `assignment`.
+Unsets the current active member from `assignment`.
 
 ### Params
-* `assignmentIds`: [ids] - assignment ids to unset a member from.
+* `assignmentIds`: [string, object] - string of assignmentIds to unset the
+   current active member from. can also be an object.
```
matthopson: "Can also be an `assignment` object."

docs/collections/assignments.md
```diff
@@ -146,38 +146,47 @@ assignmentArray.sort(teamsnap.getAssignmentSort());
...
+* `assignmentIds`: [array, object] - list of assignment ids to set the current
+   active member to. Can also be an assignment object.
+* `memberId`: [int, object] - memberId or member object of current active member.
 * `callback`: [function] - callback to be executed when the operation completes.
 
 ### Examples
 ```javascript
 // ~~~~~
-//Sets a member to an assignment.
+//Sets the current active member to multiple assignments.
```
matthopson: This is setting the specified member, not the current one (since you're passing the `memberId`).

docs/collections/assignments.md
```diff
@@ -146,38 +146,47 @@ assignmentArray.sort(teamsnap.getAssignmentSort());
...
 teamsnap.volunteerForAssignments([4, 5], 1);
+
+// ~~~~~
+//Sets the current active member to one assignment.
```
matthopson: current active / specified

docs/collections/assignments.md
```diff
@@ -189,8 +198,9 @@ teamsnap.volunteerForAssignments([4, 5], 1);
 Sends assignments emails for events to members on a team.
 
 ### Params
-* `teamId`: [id] - team id.
-* `eventIds`: [ids] - array of ids for selected events.
+* `teamId`: [int] - teamId of current team.
```
matthopson: Does it have to be the _current_ team? Could you set an assignment for any team this way? We have a few methods that use `teamId` when I don't entirely think they need to. There could be a reason for this (ie leagues) - but I don't much love it... at the very least it could be an optional param, and the `teamId` can be pulled from the `event` in most cases.

But that's a v3 concern, really.

amberwebb: It should be the current team, although there is no validation for that.

src/persistence.coffee
```diff
@@ -740,6 +742,13 @@ modifySDK = (sdk) ->
         sdk.loadMessages({messageSourceId: result.id})
         ).callback callback
 
+  wrapMethod sdk, 'optOutOfAssignments', (optOutOfAssignments) ->
```
matthopson: Is executing this command not returning the `assignment` object?

PR #138, opened 2016-07-14
Member assignments

amberwebb:
- Adds new member assignments endpoint, methods and type
- Removes validation of memberId from assignments since assignments can be unassigned (as    
   volunteer needed)
- New persistence wrapper methods


docs/collections/memberAssignments.md
```diff
@@ -0,0 +1,92 @@
+# Member Assignments
+
+## Methods
+
+- [loadMemberAssignments](#loadMemberAssignments)
+- [createAssignment](#createMemberAssignment)
```
matthopson: *createMemberAssignment

docs/collections/memberAssignments.md
```diff
@@ -0,0 +1,92 @@
...
+- [saveMemberAssignment](#saveMemberAssignment)
+- [removeMemberAssignment](#removeMemberAssignment)
```
matthopson: Just want to point out, in most cases, when we're calling a `delete` action in the SDK, we use `deleteItemName` - is there a reason for using `remove` instead?

I'm not against it, as it sounds less _harsh_, but I'd like to keep things as consistent as possible unless there are compelling reasons for the change.

amberwebb: Yeah, it felt weird using `delete` in the name since we aren't really deleting the memberAssignment, we're just removing the member from the memberAssignment.

amberwebb: I'm certainly fine with changing it to be consistent.

docs/collections/memberAssignments.md
```diff
@@ -0,0 +1,92 @@
...
+teamsnap.saveMemberAssignment(memberAsssignment);
+```
+
+
+---
+
+
+<a id="removeMemberAssignment"></a>
+## `removeMemberAssignment(memberAsssignment, callback)`
+Removes a member from a `memberAsssignment` item.
```
matthopson: Shouldn't this be more like: "Removes a member from an `assignment`." ?

docs/collections/memberAssignments.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Params
+* `memberAssignment`: [object] - memberAssignment item to remove member from.
+* `callback`: [function] - callback to be executed when the operation completes.
+
+### Examples
+``` javascript
+// ~~~~~
+// Remove member from `memberAsssignment` item.
```
matthopson: Same.

docs/collections/memberAssignments.md
```diff
@@ -0,0 +1,92 @@
...
+
+---
+
+
+<a id="removeMemberAssignment"></a>
+## `removeMemberAssignment(memberAsssignment, callback)`
+Removes a member from a `memberAsssignment` item.
+
+### Params
+* `memberAssignment`: [object] - memberAssignment item to remove member from.
```
matthopson: maybe more like: `memberAssignment` item to delete.

src/collections/assignments.coffee
```diff
@@ -59,4 +57,17 @@ exports.getAssignmentSort = (reverse) ->
       else if valueA and !valueB then -1
       else if valueA > valueB then 1
       else if valueA < valueB then -1
-      else 0
\ No newline at end of file
+      else 0
+
+exports.sendAssignmentEmails = (teamId, eventIds, message, callback) ->
+  unless teamId
```
matthopson: You could also use the @isId check here since you expect an id. This will also help validate against excluding the param. For instance if you passed `(eventIds, message)` - it would return the wrong error message.

src/sdk.coffee
```diff
@@ -135,6 +135,7 @@ TeamSnap::getDefaultSort = ->
 
 
 TeamSnap::getCollectionForItem = (item) ->
+
```
matthopson: ?

matthopson:
Missing tests.

docs/collections/memberAssignments.md
matthopson: We _should_ be deleting it. Or did I miss something? I thought a `memberAssignment` object is created ("created" in the early case) to represent the relationship between the `member` and `assignment` object... so you should be able to delete the `memberAssignment` object to unassign the `member` from the `assignment`. You're deleting the reference, not the member or assignment objects.

matthopson: This seems to support that: https://github.com/teamsnap/apiv3/blob/master/app/member_assignments/member_assignment_app.rb#L33

Or do we have `memberAssignment` objects laying around without `member` or `assignment` links on them?

docs/collections/memberAssignments.md
amberwebb: The `memberAssignment` isn't being deleted. The member is being removed.

docs/collections/memberAssignments.md
amberwebb: It could be since `memberAssignment` is a copy of `assignment`.  It makes sense this way I think, but is only telling half of the story really. There's no way for a dev to know that this is a copy of `assignments`.

docs/collections/memberAssignments.md
amberwebb: Ok. So I clearly did not understand that. Crystal clear now.

amberwebb:
@matthopson The docs are updated and I've written tests for `memberAssignments` although I'm still getting 500 errors on post and delete. JJ is working on sending back more descriptive error messages so I can debug this further. I am getting 500 and 403 errors on a few of the other tests so I wonder if the problem is my setup? 

And after looking at some of the more recent tests, we really need to go back and clean them up. Several of them are broken or not structured in the best way.

matthopson:
If you think those errors are on your end we can 😱 hero on it if you'd like.

matthopson:
👍

PR #140, opened 2016-07-18
Reload memberAssignment

amberwebb:


nathanielvelez:
👍

src/persistence.coffee
```diff
@@ -754,6 +754,12 @@ modifySDK = (sdk) ->
         sdk.loadAssignments({id: memberAssignment.assignmentId}).then -> result
         ).callback callback
 
+  wrapMethod sdk, 'saveAssignment', (saveAssignment) ->
+    (assignment, callback) ->
+      saveAssignment.call(this, assignment, callback).then((result) ->
+        sdk.loadMemberAssignments({id: result.id})
+        ).callback callback
```
matthopson: indent?

src/persistence.coffee
```diff
@@ -754,6 +754,12 @@ modifySDK = (sdk) ->
         sdk.loadAssignments({id: memberAssignment.assignmentId}).then -> result
         ).callback callback
 
+  wrapMethod sdk, 'saveAssignment', (saveAssignment) ->
```
matthopson: What about deleting?

matthopson: Also, we will probably need to ensure `memberAssignments` get removed when deleting an `event`

src/persistence.coffee
```diff
@@ -754,6 +754,12 @@ modifySDK = (sdk) ->
...
+    (assignment, callback) ->
+      saveAssignment.call(this, assignment, callback).then((result) ->
+        sdk.loadMemberAssignments({id: result.id})
```
matthopson: This should be `assignmentId`

src/persistence.coffee
amberwebb: What is there to load if an assignment is deleted than it's member assignment is deleted as well right?

matthopson: Not in memory on the front end. That's why this persistence layer exists.

Load `memberAssignments` now. Delete the associated `event`. You will likely see the same `memberAssignments.length` before and after deleting the `event`.

amberwebb:
@matthopson Please review again. I am still having an issue with unlinking the member from assignments when assignments.memberId is null.

src/persistence.coffee
```diff
@@ -740,19 +747,41 @@ modifySDK = (sdk) ->
...
+
+  wrapMethod sdk, 'deleteAssignment', (deleteAssignment) ->
+    (assignment, callback) ->
+      if assignment.memberAssignments.length
+        toRemove = []
+        assignment.memberAssignments.forEach (memberAssignment) ->
+          toRemove.push memberAssignment.member...
+          toRemove.push memberAssignment.event...
+          toRemove.push memberAssignment.assignment...
+        linking.unlinkItems toRemove, lookup
```
matthopson: When deleting an assignment, you should only have to unlink related `memberAssignment` objects - as they should be cleaned up for you by v3 when the `assignment` goes away. cc @jjmocko

amberwebb:
@matthopson Pushed changes. Please review.

Setting result.member = null did not work because the result return was for memberAssignment. I need the member on assignment to be null, or ideally unlinked.

matthopson:
Oh durr... 😛

matthopson:
👍

PR #141, opened 2016-07-28
Add sendingMemberId param to sendAssignmentEmails

amberwebb:


nathanielvelez:
I am not seeing the compiled teamsnap.js file

nathanielvelez:
Another question. Is this NJW stuff that would be using it be behind a rollout? If the old stuff that don't send a `sendingMemberId` will still be around it would feel after we update to this version of the SJK.

amberwebb:
@nathanielvelez `sendAssignmentEmails` is only available behind the assignments rollout.

nathanielvelez:
👍

PR #144, opened 2016-09-14
Add reorderAssignments method, doc and changelog updates

amberwebb:


matthopson:
Could we throw a test on this?

amberwebb:
Is there a particular reason why besides the fact that code needs testing? We haven't been keeping up with testing because of the difficulties in testing it. So I'm just curious. Should we enforce tests on all the code that gets pushed into the sdk?

matthopson:
Broken windows. We should strive to have as much coverage as humanly possible.

In this case, it should be easy... create two assignments, reverse their order, verify the order is reversed...

PR #148, opened 2016-09-29
Added bulkDeleteMessages command to messages

nathanielvelez:


src/collections/messages.coffee
```diff
@@ -17,4 +17,19 @@ exports.markMessageAsRead = (params, callback) ->
...
+    throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+      'an array of `messages` must be provided'
+
+  requestParams = { id: [] }
+  for msg in params
+    unless @isItem msg, 'message'
+      throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+        "Items in params array must have a 'message' `type`"
+    id = msg.id
+    requestParams.id.push id
```
amberwebb: The block of code from 27 to 33 is standing out to me. Why do you need to do a push here?

src/collections/messages.coffee
```diff
@@ -17,4 +17,19 @@ exports.markMessageAsRead = (params, callback) ->
...
+
+  @collections.messages.exec('bulkDelete', requestParams).callback callback
```
amberwebb: Is this command `bulkDelete` or `bulkDeleteMessages` ?

amberwebb: Looked at v3 and I see it's `bulk_delete`. Which is weird to me that it's not `bulk_delete_messages` or something like that.

src/collections/messages.coffee
```diff
@@ -17,4 +17,19 @@ exports.markMessageAsRead = (params, callback) ->
       'a message `id` or `message` object must be provided'
 
   @collections.messages.exec('markMessageAsRead', params)
-    .pop().callback callback
\ No newline at end of file
+    .pop().callback callback
+
+exports.bulkDeleteMessages = (params, callback) ->
+  if !Array.isArray params
```
amberwebb: Might be more descriptive to have the argument called `messages` instead of `params`.

nathanielvelez: 👍

src/collections/messages.coffee
nathanielvelez: @randy-girard ^

src/collections/messages.coffee
amberwebb: If v3 is expecting messageIds, then it should be messageIds instead of messages.

src/collections/messages.coffee
amberwebb: Can you get the message objects from the ids in the persistence layer?

nathanielvelez: Ok, I think i got how its done in bulkDeleteTeamMedia. You pass the object to the skd method, and in the persistence layer you extract the ids to pass it to the actual method. will do it this way instead. 👍

nathanielvelez:
@amberwebb Refactored bulkDeleteMessages to expect an id or an array of ids. The persistence layer wrapper is now expecting an array of messages or a message to do the unlinking and then sending the ids to bulkDeleteMessages. This PR is already approved but since I did some changes based on your feedback, I wouldn't mind if you re-review it and thumbs-up-it like to good old days.

src/persistence.coffee
```diff
@@ -794,6 +795,34 @@ modifySDK = (sdk) ->
         sdk.loadMessageData(params)
       ).callback callback
 
+  wrapMethod sdk, 'bulkDeleteMessages', (bulkDeleteMessages) ->
+    (messages, callback) ->
+      if Array.isArray messages
+        toRemove = messages
+      else if typeof messages is 'object' and @isItem messages, 'message'
```
matthopson: Seems like you should require an array of items.

src/persistence.coffee
```diff
@@ -794,6 +795,34 @@ modifySDK = (sdk) ->
...
+            call `bulkDeleteMessages` using the persistene layer'
+
+      messageIds = []
+      for message in toRemove
+        unless @isItem message, 'message'
+          throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+            "Items in messages array must have a 'message' `type`"
+        messageIds.push message.id
+
+      toRemove = messages
```
matthopson: This is overwriting `toRemove` from above

src/persistence.coffee
```diff
@@ -794,6 +795,34 @@ modifySDK = (sdk) ->
...
+      linking.unlinkItems toRemove, lookup
```
matthopson: What if it's an array of `id`s like in in the error message above? Unlink won't work in this case.

src/persistence.coffee
```diff
@@ -794,6 +795,34 @@ modifySDK = (sdk) ->
...
+        throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+          'Must provide an `array` of message `ids` or a `message` object to
+            call `bulkDeleteMessages` using the persistene layer'
+
+      messageIds = []
+      for message in toRemove
+        unless @isItem message, 'message'
+          throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+            "Items in messages array must have a 'message' `type`"
+        messageIds.push message.id
```
matthopson: If you don't need to do this check for each item, you can simply use `Array.map` to build this array out in one line.

src/persistence.coffee
```diff
@@ -794,6 +795,34 @@ modifySDK = (sdk) ->
...
 
+  wrapMethod sdk, 'bulkDeleteMessages', (bulkDeleteMessages) ->
+    (messages, callback) ->
+      if Array.isArray messages
+        toRemove = messages
+      else if typeof messages is 'object' and @isItem messages, 'message'
+        toRemove = [messages]
+      else
+        throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+          'Must provide an `array` of message `ids` or a `message` object to
```
matthopson: Is this accurate? You accept either an array of ids or a single object?

src/collections/messages.coffee
```diff
@@ -17,4 +17,17 @@ exports.markMessageAsRead = (params, callback) ->
       'a message `id` or `message` object must be provided'
 
   @collections.messages.exec('markMessageAsRead', params)
-    .pop().callback callback
\ No newline at end of file
+    .pop().callback callback
+
+exports.bulkDeleteMessages = (params, callback) ->
+  if Array.isArray params
+    params = id: params
```
matthopson: Do we care if this is an array of objects or ids?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "teamsnap.js",
-  "version": "1.17.0",
+  "version": "1.17.1",
```
matthopson: This should be a minor bump.

docs/collections/messages.md
```diff
@@ -50,3 +51,21 @@ var message = teamsnap.markMessageAsRead(1);
...
+---
+
+
+<a id="bulkDeleteMessages"></a>
+## `bulkDeleteMessages(params)`
+Deletes an array of `message` items.
+
+### Params
+* `params`: [array] - An array of message items.
+
```
matthopson: Callback?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+
+### October 10, 2016 // Version 1.17.1
```
matthopson: Should be a minor bump.

src/teamsnap.coffee
```diff
@@ -3,7 +3,7 @@ promises = require './promises'
 require './errors'
 
 class TeamSnap
-  version: '1.17.0'
+  version: '1.17.1'
```
matthopson: Should be minor.

src/persistence.coffee
```diff
@@ -796,10 +796,27 @@ modifySDK = (sdk) ->
...
-      toRemove = params
+    (messages, callback) ->
+      if Array.isArray messages
+        toRemove = messages
+      else if typeof messages is 'object' and @isItem messages, 'message'
+        toRemove = [messages]
+      else
+        throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+          'Must provide an `array` of message `ids` or a `message` object to
+            call `bulkDeleteMessages` using the persistene layer'
```
amberwebb: Shouldn't this check be done in the bulkDeleteMessages method? These sdk wrapper methods in persistence should just reload or unlink.

src/persistence.coffee
```diff
@@ -796,10 +796,27 @@ modifySDK = (sdk) ->
...
+
+      messageIds = []
+      for message in toRemove
+        unless @isItem message, 'message'
+          throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+            "Items in messages array must have a 'message' `type`"
+        messageIds.push message.id
```
amberwebb: Same here.

docs/collections/messages.md
amberwebb: Can you pass either an array of ids or an array of items?

src/persistence.coffee
nathanielvelez: My thinking is that it is convenient to be be able to handle both cases: an array of items or a single message item.

package.json
nathanielvelez: 👍

docs/collections/messages.md
nathanielvelez: @amberwebb right now it can handle all cases, an array of items or ids, or a single item or id.

src/collections/messages.coffee
nathanielvelez: apiv3 expects an array of ids, so if we pass an array of ids we can use it directly, but if we pass an array of objects, we should extract the ids. Will add logic to handle that case.

src/persistence.coffee
nathanielvelez: 👍

docs/collections/messages.md
amberwebb: Awesome! Can you update the docs to make that more clear?

nathanielvelez:
@matthopson @amberwebb updated with suggestions, re-review?

src/collections/messages.coffee
```diff
@@ -21,7 +21,14 @@ exports.markMessageAsRead = (params, callback) ->
...
-    params = id: params
+    if params[0]? and typeof params[0] is 'object' and
+    @isItem params[0], 'message'
+      params = id: params.map (item) -> item.id
+    else if @isId params[0]
+      params = id: params
+    else
+      throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+        'Must provide an `array` of message `ids` or `message` objects'
   else if typeof params is 'object' and @isItem params, 'message'
```
amberwebb: Should probably also check if there are no messages being passed into the method and then throw an error. Is params the best name for this? Should it be messages? Because you can only pass in message objects or message ids.

nathanielvelez: Naming it messages is a good idea. About trowing an error, My concern was that an error here will halt JS execution, so you will need to add a validation on the HTML form to be sure that you are actually passing ids or messages to delete. I can imagine an implementation where this will be called and no messages are being passed (probably by badly written code, but not sure if its worth throwing an error for that). My thinking was that such a case is better for the end user for it to fail silently. Thoughts?

src/collections/messages.coffee
```diff
@@ -21,7 +21,14 @@ exports.markMessageAsRead = (params, callback) ->
 
 exports.bulkDeleteMessages = (params, callback) ->
   if Array.isArray params
-    params = id: params
+    if params[0]? and typeof params[0] is 'object' and
+    @isItem params[0], 'message'
```
nathanielvelez: 👍

amberwebb:
It's a pattern we follow in other sdk methods. I don't think you would have to add anything in the HTML. I'm away from my Mac, but check out @request in the sdk methods.

nathanielvelez:
@amberwebb Renamed params to messages and added a checking for an empty array. Another look?

src/collections/messages.coffee
matthopson: It's pretty common in the SDK to throw an error if you're missing required fields. I think we're safe to validate on fields we know are required to prevent the request from even going out. 

In the case of badly written code, I think this makes for a better development experience.

src/collections/messages.coffee
```diff
@@ -19,22 +19,24 @@ exports.markMessageAsRead = (params, callback) ->
...
+      messages = id: messages
     else
       throw new TSArgsError 'teamsnap.bulkDeleteMessages',
         'Must provide an `array` of message `ids` or `message` objects'
-  else if typeof params is 'object' and @isItem params, 'message'
-    params = id: params.id
-  else if @isId params
-    params = id: params
+  else if typeof messages is 'object' and @isItem messages, 'message'
+    messages = id: messages.id
```
matthopson: I haven't tested this command, but I presume we don't actually need to send an array to v3 - is that correct? Or should this be Arrayified?

src/collections/messages.coffee
```diff
@@ -17,4 +17,26 @@ exports.markMessageAsRead = (params, callback) ->
...
+        'The array of messages to be deleted is empty.'
+    else if @isItem messages[0], 'message'
+      messages = id: messages.map (message) -> message.id
+    else if @isId messages[0]
+      messages = id: messages
+    else
+      throw new TSArgsError 'teamsnap.bulkDeleteMessages',
+        'Must provide an `array` of message `ids` or `message` objects'
+  else if typeof messages is 'object' and @isItem messages, 'message'
+    messages = id: messages.id
```
matthopson: I haven't tested this command, but I presume we don't actually need to send an array to v3 - is that correct? Or should this be Arrayified?

src/collections/messages.coffee
```diff
@@ -17,4 +17,26 @@ exports.markMessageAsRead = (params, callback) ->
...
+  else if @isId messages
+    messages = id: messages
```
matthopson: Same as ☝️

docs/collections/messages.md
```diff
@@ -50,3 +51,34 @@ var message = teamsnap.markMessageAsRead(1);
...
 ```
+
+---
+
+
+<a id="bulkDeleteMessages"></a>
+## `bulkDeleteMessages(params)`
+Deletes an array of `message` items.
+If you want to use the persistence layer, a message item or an array of message
+items must be passed. Persistence won't work with an id or array of ids.
```
matthopson: I would maybe change this to something like:

```
_Note: While apiv3, expects an array of ids, you must pass an array of message objects to take advantage of the persistence layer's unlinking feature. If you don't need this automatic unlinking of items, it is perfectly acceptable to just use an array of ids._
```

src/persistence.coffee
```diff
@@ -794,6 +795,22 @@ modifySDK = (sdk) ->
         sdk.loadMessageData(params)
       ).callback callback
 
+  wrapMethod sdk, 'bulkDeleteMessages', (bulkDeleteMessages) ->
+    (messages, callback) ->
+      if Array.isArray messages
+        toRemove = messages
```
matthopson: This doesn't check for an array of objects. If an array of ids are passed it would throw those in the `toRemove` array, which will fail with on `unlinkItems`

src/collections/messages.coffee
nathanielvelez: the command accepts a single id like `{id: 1}' or an array of ids `{id: [1,2,3]}`

src/collections/messages.coffee
nathanielvelez: the command accepts a single id like {id: 1}' or an array of ids{id: [1,2,3]}`

docs/collections/messages.md
nathanielvelez: 👍

nathanielvelez:
@matthopson added a check for array of objects before unlinking the message items and improved the documentation with your suggestion. Another 👀 ?

docs/collections/messages.md
```diff
@@ -50,3 +51,34 @@ var message = teamsnap.markMessageAsRead(1);
...
 ```
+
+---
+
+
+<a id="bulkDeleteMessages"></a>
+## `bulkDeleteMessages(params)`
+Deletes a single 'message' item or an array of `message` items.
+
+_Note: While apiv3 expects an array of ids, you must pass an array of message objects to take advantage of the persistence layer's unlinking feature. If you don't need this automatic unlinking of items, it is perfectly acceptable to just use an array of ids.
```
matthopson: Need to close the italics on the note with an `_`

src/persistence.coffee
```diff
@@ -794,6 +795,23 @@ modifySDK = (sdk) ->
...
+      else if typeof messages is 'object' and @isItem messages, 'message'
+        toRemove = [messages]
+      if toRemove?
+        linking.unlinkItems toRemove, lookup
+
+      bulkDeleteMessages.call(this, messages).then((result) ->
+        return result
+      ).fail((err) ->
+        linking.linkItems toRemove, lookup
```
matthopson: Should probably check `toRemove?` again here.

docs/collections/messages.md
nathanielvelez: 👍

src/persistence.coffee
nathanielvelez: 👍

nathanielvelez:
@matthopson added the extra check before linking the messages if delete fails. Also closed the missing MD tag. 👓 ?

PR #151, opened 2016-10-04
Add Divisions collection, Add bulkDeleteBroadcastEmails, divisionSearchMembers, and divisionSearchTeams methods

dmccraw:


matthopson:
So back to the convo regarding the reason for using `divisionSearchMembers` vs `searchDivisionMembers`.... It makes sense to use divisionSearchMembers, because `divisionMember` is a thing. And there is already a `loadDivisionMembers` method in the SDK.

dmccraw:
@matthopson @nathanielvelez Updated to use searchDivisionMembers and searchDivisionTeams

dmccraw:
@matthopson Updated names to `divisionLoadMembers` and `divisionLoadTeams`

PR #154, opened 2016-10-12
Reload messageData in bulkDeleteMessages persistence layer

nathanielvelez:


src/persistence.coffee
```diff
@@ -800,13 +800,23 @@ modifySDK = (sdk) ->
...
       else if typeof messages is 'object' and @isItem messages, 'message'
         toRemove = [messages]
+        teamId = messages.teamId
       if toRemove?
         linking.unlinkItems toRemove, lookup
 
       bulkDeleteMessages.call(this, messages).then((result) ->
-        return result
+        if teamId?
+          sdk.loadMessageData({teamId: teamId, messageType: 'alert,email'}).
```
amberwebb: Should you load messageData on the member or contact id instead of the teamId?

matthopson: Agree.

matthopson: Also, what's the thought behind declaring the `messageType` explicitly? I feel like if you're going to declare it, you should use the `messageType` that's being deleted.

nathanielvelez: Depending on the implementation, not sure we cannot guarantee that all messages being deleted will belong to the same member. In njw it is, notting is stopping anyone from creating something that sends messages from different contacts and members. If we wanted to be granular here, I will suggest the proper thing to do will be group all the contact and member ids of the messages being passed, and use them to load the messages data. And do the same for the messageType.

nathanielvelez:
@amberwebb @matthopson Made the query to reload messageData more granular by querying only messageData for the contactId, memberId and messageType of the deleted messages. Another 👀 ?

src/persistence.coffee
matthopson: Nothing stopping anyone from sending that request in NJW. But APIv3 should only allow the owner of the message to delete it.

src/persistence.coffee
```diff
@@ -797,16 +797,38 @@ modifySDK = (sdk) ->
...
+          if messageTypes.indexOf(message.messageType) < 0
+            messageTypes.push message.messageType
 
       bulkDeleteMessages.call(this, messages).then((result) ->
-        return result
+        if contactIds.length or memberIds.length
+          params = {}
+          if contactIds.length then params.contactId = contactIds.join()
+          if memberIds.length then params.memberId = memberIds.join()
+          if messageTypes.length then params.messageType = messageTypes.join()
```
matthopson: Compounding the search params will only scope the request further, not broaden it.

nathanielvelez: Yep tested it. I liked the idea of doing this reloading on njw where he have more context for the request, but it made sense to do so in in the persistence layer instead. Initially I was reloading al messageData for the team since that was a simpler query, but the idea here was to reduce the scope, so we are not loading messageData for the whole team. In this case because we are testing for the existence of the params on the message before adding it to the query, the scoping works in this case (we need to query for either member or contact id depending message).

src/persistence.coffee
```diff
@@ -797,20 +797,32 @@ modifySDK = (sdk) ->
...
+            messageTypes.push message.messageType
 
       bulkDeleteMessages.call(this, messages).then((result) ->
-        if teamId?
-          sdk.loadMessageData({teamId: teamId, messageType: 'alert,email'}).
-          then((result) ->
+        if contactIds.length or memberIds.length
+          params = {}
+          if contactIds.length then params.contactId = contactIds.join()
+          if memberIds.length then params.memberId = memberIds.join()
```
nathanielvelez: We need to be sure that the messages deleted were from a contact or a member, so we to filter the query on the params that the messages contain. For example querying {memberId: 1, contactId: ''} will return no messageData. It can be done in other probably, but not that much simpler since we don't have the full context of njw when doing this reloading from the sdk. Search for just teamId would have been simpler, but it made sense to scope it as mentioned on previews feedback.

src/persistence.coffee
```diff
@@ -797,20 +797,32 @@ modifySDK = (sdk) ->
...
-        teamId = messages.teamId
       if toRemove?
         linking.unlinkItems toRemove, lookup
+        for message in toRemove
+          if message.contactId? and contactIds.indexOf(message.contactId) < 0
+            contactIds.push message.contactId
+          if message.memberId? and memberIds.indexOf(message.memberId) < 0
+            memberIds.push message.memberId
+          if messageTypes.indexOf(message.messageType) < 0
+            messageTypes.push message.messageType
```
nathanielvelez: Doing more testing, this query is kind of cranky in apiv3. If you don't specify the messageType it doesn't return anything. And because other quirks on that this endpoint returns, the best implementation is to unlink all message data and reload both alerts and messages here. (if there are no unread messages, this query also return 0, which breaks the auto unlinking if going from 1 unread to 0)

nathanielvelez:
@amberwebb @matthopson Updated with suggestions. 
A few notes:
- apiv3 requires that you specify messageType for it to return anything with this enpoint
- If you delete a message when going from 1 unread to 0, apiv3 returns nothing, causing the unread badge to never reach 0, Thus the messageData items must be unlinked in njw to work around this. because of this, we need to reload message data both emails and alerts.
- Since we are reloading messageData from the skd and not reloading messageData for the whole team, we need to see if the messages deleted belong to a member or a contact. To simplify the whole thing, this query assumes that are messages being deleted belong to the same person.
- I am also sending the teamId to prevent loading messagesData form other teams. With the auth token apiv3 filters those by team, but in case the token gets messed up as it happened recently, this will prevent data leaking from other teams.

Anohter 👓 ?

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### Oct 12, 2016 // Version 1.19.1
```
matthopson: Today is the 13th

PR #157, opened 2016-11-17
CreateBulkAssignments

amberwebb:
Depends on https://github.com/teamsnap/apiv3/pull/2570


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "teamsnap.js",
-  "version": "1.20.0",
+  "version": "1.20.1",
```
ryanwi: Would a new method just be a patch update? Seems like it might be worth a minor bump?

src/teamsnap.coffee
```diff
@@ -3,7 +3,7 @@ promises = require './promises'
 require './errors'
 
 class TeamSnap
-  version: '1.20.0'
+  version: '1.20.1'
```
ryanwi: same version question

amberwebb: Yep. You are right. Thanks!

src/collections/assignments.coffee
```diff
@@ -86,3 +86,24 @@ exports.reorderAssignments = (eventId, assignmentIds, callback) ->
...
+    throw new TSArgsError 'teamsnap.createBulkAssignments',
+    " `eventSet` must be provided."
+  unless description.trim()
+    return @reject 'You must provide a description for the assignments.',
+      'description', callback
+  unless teamId
+    throw new TSArgsError 'teamsnap.createBulkAssignments', '`teamId` must be
+    provided.'
+  if @isItem teamId, 'team'
+    teamId: teamId.team
```
matthopson: should be `teamId.id`

src/collections/assignments.coffee
```diff
@@ -86,3 +86,24 @@ exports.reorderAssignments = (eventId, assignmentIds, callback) ->
   params = eventId: eventId, sortedIds: assignmentIds
   @collections.assignments.exec('reorderAssignments', params)
     .callback callback
+
```
matthopson: What do you think about putting those `eventSet` strings into the SDK?

You could do something like:
```coffeescript
eventSets = [
  'future_events',
  'future_games',
  'future_games_and_events'
]
exports.ASSIGNMENTS_EVENT_SETS = eventSets
```

This way they can be referenced inside the app as well (rather than hard-code them) - and in your validation below you could do something like:

```coffeescript
unless eventSets.indexOf(eventSet) > -1
  throw new TSArgsError 'teamsnap.createBulkAssignments',
    " `eventSet` must be one of the following: " + eventSets.toString() +."
```

src/persistence.coffee
```diff
@@ -832,6 +833,13 @@ modifySDK = (sdk) ->
         err
       ).callback callback
 
+  wrapMethod sdk, 'createBulkAssignments', (createBulkAssignments) ->
+    (assignments, callback) ->
+      createBulkAssignments.call(this, assignments, callback).then((result) ->
+        assignmentIds = result.map (assignment) -> assignment.id
```
matthopson: So is the result of the bulk create just a list of ids? Doesn't return any objects?

src/collections/assignments.coffee
amberwebb: I like that idea. Where would you suggest to add that? I would think it could just be added to assignments.coffee. I don't see this being referenced anywhere else in the sdk.

matthopson: Yup, just throw it at the top of assignments.coffee. We do something similar in availabilites.coffee

amberwebb:

PR #158, opened 2016-11-19
add bulkDeleteMembers, moveMemberToTeam, divisionAdvancedLoadMembers, and memberRegistrationSignups, registrationForms collections

dmccraw:
MoveMember


src/collections/members.coffee
```diff
@@ -105,6 +105,13 @@ exports.divisionLoadMembers = (params, callback) ->
   @collections.members.queryItems('divisionSearch', params, callback)
 
 
+exports.divisionAdvancedLoadMembers = (params, callback) ->
```
matthopson: Does this query only work for divisions?

src/collections/members.coffee
```diff
@@ -175,3 +182,55 @@ exports.loadImportableMembers = (userId, includeArchivedTeams, callback) ->
   params = userId: userId, includeArchivedTeams: includeArchivedTeams
 
   @collections.members.queryItems 'importableMembers', params, callback
+
+exports.bulkDeleteMembers = (members, callback) ->
+  if Array.isArray members
+    if members.length is 0
+      throw new TSArgsError 'teamsnap.bulkDeleteMembers',
+        'The array of members to be deleted is empty.'
+    else if @isItem members[0], 'member'
```
matthopson: Do you think it would be better to do a `members.every()` and check for `member` type?

src/collections/members.coffee
```diff
@@ -175,3 +182,55 @@ exports.loadImportableMembers = (userId, includeArchivedTeams, callback) ->
...
+        'Must provide an `array` of member `ids` or `member` objects'
+  else if typeof members is 'object' and @isItem members, 'member'
+    members = memberId: members.id
+  else if @isId members
+    members = memberId: members
+  else
+    throw new TSArgsError 'teamsnap.bulkDeleteMembers',
+      'Must provide an `array` of member `ids`, an `id` or a `member` object'
+
+  @collections.members.exec('bulkDelete', members).callback callback
```
matthopson: What happens when there's an error deleting one of the `member` items?

src/collections/members.coffee
```diff
@@ -175,3 +182,55 @@ exports.loadImportableMembers = (userId, includeArchivedTeams, callback) ->
...
+    else if @isItem members[0], 'member'
+      members = memberId: members.map((member) -> member.id)
+    else if @isId members[0]
+      members = memberId: members
+    else
+      throw new TSArgsError 'teamsnap.bulkDeleteMembers',
+        'Must provide an `array` of member `ids` or `member` objects'
+  else if typeof members is 'object' and @isItem members, 'member'
+    members = memberId: members.id
+  else if @isId members
```
matthopson: I wonder if we should require `member` items instead of ids, just so we can validate. Bulk actions make me nervous, but maybe it's not that big of a deal?

src/collections/members.coffee
```diff
@@ -175,3 +182,55 @@ exports.loadImportableMembers = (userId, includeArchivedTeams, callback) ->
   params = userId: userId, includeArchivedTeams: includeArchivedTeams
 
   @collections.members.queryItems 'importableMembers', params, callback
+
+exports.bulkDeleteMembers = (members, callback) ->
```
matthopson: I don't see a wrapper for the persistence layer for this.

src/collections/members.coffee
```diff
@@ -175,3 +182,55 @@ exports.loadImportableMembers = (userId, includeArchivedTeams, callback) ->
...
+    members = memberId: members.id
+  else if @isId members
+    members = memberId: members
+  else
+    throw new TSArgsError 'teamsnap.bulkDeleteMembers',
+      'Must provide an `array` of member `ids`, an `id` or a `member` object'
+
+  @collections.members.exec('bulkDelete', members).callback callback
+
+exports.moveMember = (params, callback) ->
```
matthopson: I feel like we should call this something more descriptive, like `memberTransferTeams` or something along those lines. "move" doesn't really sound like anything specific.

src/collections/members.coffee
```diff
@@ -175,3 +182,55 @@ exports.loadImportableMembers = (userId, includeArchivedTeams, callback) ->
...
+
+  if Array.isArray params.memberId
+    if params.memberId.length is 0
+      throw new TSArgsError 'teamsnap.moveMember',
+        'The array of memberId in params is empty.'
+    else if @isItem params.memberId[0], 'member'
+      params.memberId = params.memberId.map((member) -> member.id)
+  else if @isItem params.memberId, 'member'
+    params.memberId = params.memberId.id
+  else if @isId params.memberId
```
matthopson: Same as bulkDelete comments - should we only allow objects?

src/collections/members.coffee
```diff
@@ -175,3 +182,55 @@ exports.loadImportableMembers = (userId, includeArchivedTeams, callback) ->
...
+    `memberId`'
+  unless params.divisionId
+    throw new TSArgsError 'teamsnap.moveMember', 'params must include
+    `divisionId`'
+
+  if Array.isArray params.memberId
+    if params.memberId.length is 0
+      throw new TSArgsError 'teamsnap.moveMember',
+        'The array of memberId in params is empty.'
+    else if @isItem params.memberId[0], 'member'
```
matthopson: Same as bulkDelete comments - should we validate `every()` item in the array?

src/collections/members.coffee
```diff
@@ -175,3 +182,55 @@ exports.loadImportableMembers = (userId, includeArchivedTeams, callback) ->
...
+    throw new TSArgsError 'teamsnap.moveMember',
+      'Must provide an `array` of member `ids` or `member` objects for memberId'
+
+  if @isItem params.divisionId, 'division'
+    params.divisionId = params.divisionId.id
+
+  if @isItem params.teamId, 'team'
+    params.teamId = params.teamId.id
+
+  @collections.members.exec('moveMember', params).callback callback
```
matthopson: Ah, I see, the command is called `moveMember`.... I think I'd still like to see the SDK use a more descriptive name.

docs/collections/members.md
```diff
@@ -353,3 +355,60 @@ teamsnap.divisionLoadMembers({divisionId: 1, isActive: true});
...
+teamsnap.bulkDeleteMembers(1);
+```
+
+---
+
+
+<a id="moveMember"></a>
+## `moveMember(params, callback)`
+Moves a single 'memberId' item or an array of `memberId` items to either divisionID or teamId.
+teamId must be inside the divisionId.
```
matthopson: Does that mean root `divisionId`? Or Can you only move members between teams at the lowest division level?

docs/collections/members.md
```diff
@@ -353,3 +355,60 @@ teamsnap.divisionLoadMembers({divisionId: 1, isActive: true});
...
+teamsnap.bulkDeleteMembers(members);
+
+// Deletes a single member by id.
+teamsnap.bulkDeleteMembers(1);
+```
+
+---
+
+
+<a id="moveMember"></a>
```
matthopson: See comments re: naming

docs/collections/members.md
dmccraw: This means the teams must be in or under that divisionId.

matthopson: So you can't move a member to a sibling division?

src/collections/members.coffee
dmccraw: Yeah this is specifically calling the `members/advanced_division_search`

src/collections/members.coffee
dmccraw: if we require member items then it's much easier for persistance

src/collections/members.coffee
dmccraw: Do we have any examples of this kind of check anywhere?

src/persistence.coffee
```diff
@@ -841,6 +843,30 @@ modifySDK = (sdk) ->
...
+        toRemove.push member.trackedItemStatuses...
+        toRemove.push member.memberPayments...
+        toRemove.push member.memberStatistics...
+        toRemove.push member.statisticData...
+        toRemove.push member.memberAssignments...
+
+      linking.unlinkItems toRemove, lookup
+      bulkDeleteMembers.call(this, members, callback).then((result) ->
+        return result
+        ).fail((err) ->
```
matthopson: Just curious about how the bulk command actually works with error handling. If it has an error deleting a member in v3, does it roll back the whole transaction and throw an error? (assuming this is the case). --- or does it still delete some things and hoarf on the rest?

dmccraw: It will throw an error just like if one member was deleted and failed and it won't delete any of them.

dmccraw:
@matthopson Updated

PR #163, opened 2017-01-11
Use latest brunch

matthopson:


package.json
```diff
@@ -1,6 +1,10 @@
 {
   "name": "teamsnap.js",
-  "version": "1.23.0",
+  "version": "1.22.0",
+  "engines": {
```
amberwebb: Why did the teamsnap.js version number go from 1.23.0  to 1.22.0 ?

matthopson: Must not have pulled master (thought I did).

matthopson:
@amberwebb Not sure why that version was off - rebased and it still stayed off. I verified the other changes were there from that PR and fixed the version.

PR #164, opened 2017-02-23
Add sendInvites to importMembersFromTeam

amberwebb:


src/collections/members.coffee
```diff
@@ -160,7 +160,7 @@ exports.canEditItem = (member, team, item) ->
   else
     item.links.member.href is member.href
 
-exports.importMembersFromTeam = (memberIds, teamId, callback) ->
+exports.importMembersFromTeam = (memberIds, teamId, sendInvites, callback) ->
```
nathanielvelez: where is `sendInvites` being used in this script?

amberwebb: @nathanielvelez I just added it in the params.

PR #165, opened 2017-02-24
Add sendInvites to importMembersFromTeam in persistence

amberwebb:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### February 27, 2017 // Version 1.23.2
+- Adds `sendInvites` to `importMembersFromTeam` method in persistence layer.
+
+---
+
 ### February 24, 2017 // Version 1.23.1
 - Adds `sendInvites` to `importMembersFromTeam`.
```
knappe: How is the above message different than this one?

knappe: Are we really fixing a bug here?

amberwebb: We have a `importMembersFromTeam` method in the members collection. We also have a method in the persistence layer that wraps `importMembersFromTeam` method. That wrapper method needs the `sendInvites` parameter, or we do not get the `sendInvites` field in the request. So yes, this is fixing a bug.

And I've added "method" to the previous changelog message for clarity.

amberwebb:
@knappe Please review again.

knappe:
Can we mention that this was a bug fix? We want consumers of the sdk to be aware that we've fixed a bug they may have stumbled across.

amberwebb:
@knappe I've updated the message to reflect that it is a bug fix. Are you good with this?

PR #167, opened 2017-03-11
First draft - event emitter on sdk request

matthopson:
This is a first-draft of the event emitter in the SDK's request cycle.
Basic usage would be something like:

```javascript
document.addEventListener('teamsnap-sdk-event', (sdkEvent) => {
  const { collection, method, refIds } = sdkEvent.detail
  switch (method) {
    case 'get':
      // Now you know some new ids were loaded
      console.log(`The following IDs were loaded on ${collection} : ${refIds.toString()}`
    case 'post':
      // Now you know to load up these new ids
      console.log(`The following IDs were created on ${collection} : ${refIds.toString()}`
      break  
    case 'put':
      // Now you know to update these existing ids
      console.log(`The following IDs were updated on ${collection} : ${refIds.toString()}`
      break
    case 'delete':
      // Now you know to remove these existing ids
      console.log(`The following IDs were deleted on ${collection} : ${refIds.toString()}`
      break
  }
})
```

You can trigger other data-dependency events based on this (like triggering messages in Redux)

Feel free to beat it up / leave some feedback.


matthopson:
Added `get` method back in. I didn't think it would be as useful (and probably kinda noisy), but I think it could be really useful.

PR #171, opened 2017-04-07
Reload memberEmailAddresses and contactEmailAddresses in disableMember method

amberwebb:


lib/teamsnap.js
```diff
@@ -5838,6 +5838,18 @@ modifySDK = function(sdk) {
       }).callback(callback);
     };
   });
+  return wrapMethod(sdk, 'disableMember', function(disableMember) {
+    return function(memberId, callback) {
+      return disableMember.call(this, memberId, callback).then(function(result) {
+        sdk.loadMemberEmailAddresses({
```
dmccraw: Is there a timing issue here? You are returning `loadContactEmailAddresses` so that promise will be handled the the caller of disableMember but does it matter that `loadMemberEmailAddresses` is finished before this is returned? Could `.callback(callback)` get called before `loadMemberEmailAddresses` is finished and does it matter?

src/persistence.coffee
```diff
@@ -867,6 +869,13 @@ modifySDK = (sdk) ->
           err
       ).callback callback
 
+  wrapMethod sdk, 'disableMember', (disableMember) ->
+    (memberId, callback) ->
+      disableMember.call(this, memberId, callback).then((result) ->
```
matthopson: Are we not concerned with phone numbers?

src/teamsnap.coffee
```diff
@@ -3,7 +3,7 @@ promises = require './promises'
 require './errors'
 
 class TeamSnap
-  version: '1.25.3'
+  version: '1.25.4'
```
matthopson: Should probably be a minor bump, since this directly affects the API.

lib/teamsnap.js
amberwebb: I don't think it matters. I just need to reload the email addresses for `receivesTeamEmails` to be set to false once `disableMember` is called.

src/persistence.coffee
amberwebb: Ah, good question. I think we should be concerned about that. Would require an api change though, so let me get verification on that. @ryanwi What do you think?

matthopson:
What is `disableMember` actually doing in v3? It's a little ambiguous - I assume it's removing their ability to login / nulling their invitation?

dmccraw:
apiv3
```
          contact.with(
            :user_id => nil,
            :invitation_code => nil,
            :invitation_declined => nil,
            :allow_shared_access => false
          )
        member.with(
          :user_id => nil,
          :invitation_code => nil,
          :invitation_declined => nil
        )
```

matthopson:
It worried me a little since this is a case that will be directly affected by the eventEmitter - I need to start on v2 of this to cover commands and queries.

src/teamsnap.coffee
amberwebb: How does it directly affect the api? I'm just reloading the objects. Historically in the changelog we have made persistence changes to be patch updates.

matthopson: I _think_ that's a mistake, then - others should weigh in, but in my opinion, if calling a method has a different response than it used to, that is an API change - and thus should be a minor bump.

amberwebb: @dmccraw @nathanielvelez Thoughts on this? We should get this settled for future updates.

dmccraw: Here is apiv3's version https://github.com/teamsnap/kb/issues/4
Not sure the best way to do boundaries in the sdk. I do like matt's logic that this changes the results from an api call which is similar to apiv3's serializer change.

src/persistence.coffee
amberwebb: Yep. Mike O is going to work on the api changes. Will need to reload phone numbers and member prefs.

src/teamsnap.coffee
amberwebb: Yep okay. Makes sense. http://semver.org/

nathanielvelez: I kind of agree with Matt. My app will behave different with this change if I am using this method and that might be unexpected.

nathanielvelez:
@amberwebb @matthopson @dmccraw @Ollo @mikereinmiller I am not saying this this to stop reloading stuff from the sdk right away, but just as a conversation starter (that we can pick up latter). This pattern of reloading data thru the persistence layer goes a bit against the functional pattern we are trying to move NJW to and that we currently use in LFE. 

In my mind, reloading data should be handled by a redux action using an async middleware. Having all those callbacks happening automatically out of our control I think is kind of an anti-pattern in the react-redux world. I have been using `teamsnap.doWathever.oldMethod.call(teamsnap, payload)` all over the place in my messages project just to avoid reloading data outside of my component life cycle. It is a bit tedious, but I have been able to get really good results in terms of performance and preventing weird stuff from happening by having granular control of when data is being loaded.

I think we should start thinking how our SDK might need to evolve in a post chip / controller world. Do we start transforming it into a more functional beast, or do we continue updating it for a more controller / general purpose world?

PR #175, opened 2017-05-10
Convert persistentUuid value to string

matthopson:
- JavaScript can't properly represent integers over a certain size. We have a specific case in `persistentUuid` that is failing.
- It must be addressed at the request response level before any vars are assigned or computation is done (so before `JSON.parse` can even happen).
- This does a `replace` on that field specifically.
- Future implementation may replace any integer that's too long with a string rather than just this field.
- Versioning looks off at the moment, as I'm waiting on a few other branches to go in first.


amberwebb:
When is `persistenUuid` used? What is the use case for doing this?

coreycondardo:
@amberwebb current we are using this in the Hamm app - it is Club & League billing. We parse this value and post it to Salesforce. It is our key identifier to connect 3rd parties to the root league. Leagues have an ID that is constantly changing because of archives. When a league is archived it generates a new ID. The persistent_uuid never changes, so it is a sure fire way to get the latest active root division. We discovered this bug when creating new leagues as JS can't handle parsing the bigint. It was rounding the ID which poses problems since salesforce receives the wrong ID and stores it, totally ruining all data integrity.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### May 9, 2017 // Version 1.29.0
+- Parse `persistentUuid` field as string in response from v3. While this is _technically_ a potentially breaking change for some, this field is mostly unusable since it is a `bigint` (and may produce values unsupported by  JS).
+
```
dmccraw: extra space between by and JS

amberwebb:
Ok, this is probably a silly question. Is this change to specific to `persistentUuid` ? Should it be made more generic to encompass converting all bigints?

coreycondardo:
@amberwebb probably a question for @matthopson I'm not really sure how many we have.

matthopson:
@amberwebb I agree, that we should try and cover it across the board. I really wanted to get this fix in for now just to see how it goes. It's _technically_ a breaking change, since it's modifying the response in the API - which makes me cringe so hard.... BUT... in this case, this field is completely useless for anyone trying to use it, so I'm not _too_ upset about breaking the API so that it actually _works_.

I left a note to this effect in the file that's doing the magic.

PR #177, opened 2017-05-16
handle _teamsnapReturnCollection flag

matthopson:
- When calling a method that uses `loadItems`, allow `_teamsnapReturnCollection` param to be passed to return the entire `collection` response rather than just the `items`. 
- This is extremely useful for getting pagination information.
- The param will be ignored when the final request is built (because magic) and not passed along to v3.


matthopson:
@amberwebb @nathanielvelez Would like to both your thoughts on this. Probably something you should be aware of at the very least.

I tried to go about this with a wrapper method, but there's just no good way to intercept the response when creating a new request. I think passing a param is a little dirty. It _will_ get removed when the request is serialized, but still a little smelly... just may not be possible to do it another way without a lot of a lot of rework.

nathanielvelez:
@matthopson Can you provide an example of how this would be used in the app?

If what you are trying to do is what I imagine, what about creating a `loadPagedItems` method instead of passing the extra parameter parameter? This way we ensure that the `loadItems` always return an array of items and `loadPagedItems` will always return an object with the loaded items and the parameters for handling pagination.

amberwebb:
@matthopson Is this for pagination? Are there any other use cases for doing this?

matthopson:
@nathanielvelez @amberwebb Yup. So normally if you did:
`teamsnap.loadMembers({teamId: 5})`
You'd get back: `[Items]`

So with this you could do: `teamsnap.loadMembers({teamId: 5, _teamsnapReturnCollection: true})` and get back `[collection]` - which contains `[items]` as well as pagination info, etc... so in a real world usage:

```teamsnap.loadMembers({teamId: 5, pageSize: 10, number: 1, _teamsnapReturnCollection: true}).then(function(collection) {
  var items = collection.items;
  var totalItems = collection.page_info.total_items
});
```

amberwebb:
I like it! Sold! :gavel:

PR #178, opened 2017-05-19
WIP LIG-1543 set header for ghost contact

amberwebb:
As I understand the auth process, `teamsnap.auth()` is called from njw with the auth token that is sent down from classic and attached to the window object (scary, btw). `Teamsnap.auth` sets the authentication header and it clones sdkRequest method, which set the accept params for the collection json. Headers are set and we move on.

I'm not completely happy with the way I have set the feature header. While it does work and depends on the options passed in `teamsnap.auth()` in njw, I think maybe this could be extracted into it's own method and then called here in `teamsnap.auth`? But really `teamsnap.auth()` isn't true, because this function is doing more than just authorizing. It's also setting the accept params for collection.json AND now it's setting feature headers that are passed down from njw.

Thoughts on this?


amberwebb:
Companion PR https://github.com/teamsnap/nextjenn-web/pull/2544

src/auth.coffee
```diff
@@ -77,7 +77,7 @@ generateUrl = (endpoint, params) ->
 
 
 # Return an authed SDK object for use with a single authorized user
-TeamSnap::auth = (token) ->
+TeamSnap::auth = (token, optionalRequestHeaders) ->
```
matthopson: Do you think this param would be better if it was simply `options`, and you could pass `{ optionalRequestHeaders: [] }` to it? I think this would allow us to do more things down the road here and not limit ourselves to _just_ passing those headers.

src/auth.coffee
```diff
@@ -104,6 +104,11 @@ TeamSnap::auth = (token) ->
 
   @request.hook (xhr) ->
     xhr.setRequestHeader 'Authorization', 'Bearer ' + token
+
+  if options and options.length
```
matthopson: In this case, it looks like you should be checking for `optionalRequestHeaders` instead of `options` - but I think I like the second param to be an `options` object with `optionalRequestHeaders` as a property (see other comment).

src/auth.coffee
```diff
@@ -104,6 +104,11 @@ TeamSnap::auth = (token) ->
...
+    for option in optionalRequestHeaders
+      @request.hook (xhr) ->
+        xhr.setRequestHeader option.name, option.value
```
matthopson: I think I'd like to see `header.name` and `header.value` so it reads a little better.

matthopson: Also, were you not able to conditionally apply this setRequestHeader loop above (line 105) where we're already setting header?

amberwebb: I did not try that. I was trying to keep auth headers and optional headers separated, but I guess it would be better to do it there since we have already open up the request.

matthopson: Yeah, I don't see any harm in including it there. I think it will be plenty obvious what's going on.

src/auth.coffee
amberwebb: :doh: Yes, agreed.

amberwebb:

src/auth.coffee
```diff
@@ -104,6 +104,10 @@ TeamSnap::auth = (token) ->
 
   @request.hook (xhr) ->
     xhr.setRequestHeader 'Authorization', 'Bearer ' + token
+    
+    if options.optionalRequestHeaders and options.optionalRequestHeaders.length
```
matthopson: You could do:
`if options?.optionalRequestHeaders?.length`
to simplify.

amberwebb:
@matthopson Done

matthopson:
I like it. Build that sucker out and ping me.

amberwebb:
@matthopson Built and updated.

PR #181, opened 2017-06-05
Creates method for inviting contact email addresses and reloads contact email addresses

amberwebb:


src/collections/contactEmailAddresses.coffee
```diff
@@ -34,3 +34,20 @@ exports.deleteContactEmailAddress = (contactEmailAddress, callback) ->
       '`contactEmailAddress` must be provided'
 
   @deleteItem contactEmailAddress, callback
+
+
+exports.inviteContactEmailAddresses = (options, callback) ->
+  unless options.teamId
```
matthopson: We should first check for params object before evaluating its properties to prevent a reference error. Something like:
```coffeescript
unless params and typeof params is 'object'
  throw new TSArgsError 'teamsnap.inviteContactEmailAddresses', 'must provide 
  a params object'
```

src/collections/contactEmailAddresses.coffee
```diff
@@ -34,3 +34,20 @@ exports.deleteContactEmailAddress = (contactEmailAddress, callback) ->
...
+  unless options.memberId
+    throw new TSArgsError 'teamsnap.inviteContactEmailAddresses',
+      'options.memberId is required.'
+
+  unless options.contactEmailAddressIds
+    throw new TSArgsError 'teamsnap.inviteContactEmailAddresses',
+      'options.contactEmailAddressIds is required.'
+
+  @collections.contactEmailAddresses
+    .exec('invite', options).pop().callback callback
```
matthopson: We're only returning one item?

docs/collections/contact_email_addresses.md
```diff
@@ -117,3 +118,39 @@ teamsnap.saveContactEmailAddress(contactEmailAddress).then(function(){
   });
 });
 ```
+<a id="inviteContactEmailAddresses"></a>
+## `inviteContactEmailAddresses(params, callback)`
+Invite contactEmailAddresses to a team
+
+### Params
+* `options`: [object] - Object that contains additional data:
+  * `contactEmailAddressIds`: [string] A comma separated list of ids
```
amberwebb: No, it takes a string.

docs/collections/contact_email_addresses.md
```diff
@@ -117,3 +118,39 @@ teamsnap.saveContactEmailAddress(contactEmailAddress).then(function(){
...
+  contactId: 1,
+  email: 'contact@example.com',
+  receivesTeamEmails: true
+});
+
+var options = {
+  contactEmailAddressIds: [1, 2, 3]
+  teamId: 1,
+  memberId: 2,
+  notifyAsMemberId: 1,
```
amberwebb: `notifyAsMemberId` does need to be removed. This is a copy pasta error from memberEmailAddresses. :)

But introduction text does not need to be a part of this, as we are no longer sending custom messages with the email. The `introductionText` param needs to be removed from the endpoint.

@ryanwi Can you confirm the above? Or am I totally off base?

amberwebb:

docs/collections/contact_email_addresses.md
matthopson: In the past, we've used an array for these.
https://github.com/teamsnap/teamsnap-javascript-sdk/blob/2be02f2fe6b7b5fcfa29b48e70fedeaaba6b6995/docs/collections/broadcast_emails.md#params-4
https://github.com/teamsnap/teamsnap-javascript-sdk/blob/483b02886e8574a4dd3df62d742b01193a340eac/docs/collections/statistic_groups.md#params-4
https://github.com/teamsnap/teamsnap-javascript-sdk/blob/483b02886e8574a4dd3df62d742b01193a340eac/docs/collections/statistics.md#params-4
https://github.com/teamsnap/teamsnap-javascript-sdk/blob/5e20b83628cf59f150e5adebe48010d351b96b33/docs/collections/assignments.md#params-5

amberwebb: @ryanwi Can you verify that `contactEmailAddressIds` should be a string and not an array?

ryanwi: we can do arrays, but was using what we had for member email addresses, which is stings

https://github.com/teamsnap/teamsnap-javascript-sdk/blob/5e20b83628cf59f150e5adebe48010d351b96b33/docs/collections/member_email_addresses.md#inviteMemberEmailAddresses

why this is a unicorn using strings instead of an array is a good question, I'll see what I can dig up.

knappe: We're inconsistent, but we ought to expect a comma separated string.  The collection json spec: https://github.com/collection-json/extensions/blob/master/command-templates.md#10-command-templates

PR #184, opened 2017-06-26
Remove facebookPage from types

amberwebb:
Facebook endpoint in v3 is deprecated and the UI code has been removed from the web app. This PR removes `facebookPage` from types to prevent it from being bulk loaded into the app and causing errors.


nathanielvelez:
should you remove: 
- `src/collections/facebookPages.coffee`
- https://github.com/teamsnap/teamsnap-javascript-sdk/blob/1df0596144eb21fa8f3e7cb64b708967385d9571/src/sdk.coffee#L192  
- line 26 on `docs/collections.md`

???

amberwebb:
@nathanielvelez I believe we should still require the collection. The endpoint is deprecated but not completely removed until v4 of the api.

PR #185, opened 2017-07-24
Reload contacts and contactEmailAddresses on invite

amberwebb:


amberwebb:
Maybe linking to the PR is something we can require in future versions of the sdk. That would need to be decided upon by the guild and can be discussed at our next meeting.

PR #190, opened 2017-08-09
MNG-2741 - Adds registrationLineItems and registrationLineItemOptions

Ollo:
<https://teamsnap.atlassian.net/browse/MNG-2741>


src/types.coffee
```diff
@@ -58,6 +58,8 @@ types = [
   'paymentNote'
   'plan'
   'registrationForm'
+  'registrationLineItemOptions'
+  'registrationLineItems'
```
matthopson: These should be singular types:
```
registrationLineItemOption
registrationLineItem
```

matthopson:
It may seem like re-arranging deck chairs on the Titanic, but we're also missing docs for these. I know we're going to move on beyond the SDK, but I'd like to keep it complete. If you're going to be adding additional methods for these collections, you can drop the docs in then (they're pretty simple).

src/collections/registrationLineItemOptions.coffee
```diff
@@ -0,0 +1,6 @@
+exports.loadRegistrationLineItemOptions = (params, callback) ->
+  unless params and typeof params is 'object'
+    throw new TSArgsError 'teamsnap.registrationLineItemOptions',
+      'must provide query parameters'
+
+  @loadItems 'registrationLineItemOptions', params, callback
```
matthopson: Should be calling `loadItems` against the singular type.

src/collections/registrationLineItems.coffee
```diff
@@ -0,0 +1,6 @@
+exports.loadRegistrationLineItems = (params, callback) ->
+  unless params and typeof params is 'object'
+    throw new TSArgsError 'teamsnap.registrationLineItems',
+      'must provide query parameters'
+
+  @loadItems 'registrationLineItems', params, callback
```
matthopson: Same here - should be `@loadItems registrationLineItem`

src/collections/registrationLineItemOptions.coffee
```diff
@@ -0,0 +1,6 @@
+exports.loadRegistrationLineItemOptions = (params, callback) ->
+  unless params and typeof params is 'object'
+    throw new TSArgsError 'teamsnap.registrationLineItemOptions',
```
matthopson: Should include the method name in the error:
`throw new TSArgsError 'teamsnap.loadRegistrationLineItemOptions'`

src/collections/registrationLineItems.coffee
```diff
@@ -0,0 +1,6 @@
+exports.loadRegistrationLineItems = (params, callback) ->
+  unless params and typeof params is 'object'
+    throw new TSArgsError 'teamsnap.registrationLineItems',
```
matthopson: method name here

src/collections/registrationLineItemOptions.coffee
```diff
@@ -0,0 +1,6 @@
+exports.loadRegistrationLineItemOptions = (params, callback) ->
+  unless params and typeof params is 'object'
```
matthopson: There may be a few cases where we've diverted from this - but we should allow a `teamId` in addition to an object with params. That would look more like:
```javascript
exports.loadRegistrationLineItemOptions = (params, callback) ->
  if @isId params
    params = teamId: params
  else unless params and typeof params is 'object'
    throw new TSArgsError 'teamsnap.loadRegistrationLineItemOptions', 'must provide a teamId or
      query parameters'

@loadItems 'registrationLineItemOption', params, callback
```

matthopson: This allows you to simply do: `loadRegistrationLineItemOptions(teamId)` to load for a given team.

dmccraw: Matt, there isn't any connection to a team for a RegistrationLIneItem or a RegistrationLineItemOption. These are for registrants going through registration before they are ever assigned to a team. There is a division associated with them.

matthopson: Ah. Good thing I provisioned with "There may be a few cases where we've diverted from this"

Thanks Dustin!

Ollo:
@matthopson updated those to remove the `teamId` check

PR #191, opened 2017-08-14
Adds types to sdk.coffee

Ollo:
@matthopson I've added the types to the sdk file as you suggested in an attempt to fix loading them in LFE. Please advise if there is anything else I missed in the first pass.


Ollo:
@matthopson Thanks! Can you assist with publishing the changes :)

PR #197, opened 2017-09-22
Oev 20 send email validation

matthopson:
Adds `sendEmailValidation` command  to `User`


src/collections/users.coffee
```diff
@@ -21,4 +21,8 @@ exports.saveUser = (user, callback) ->
   unless user.email?.trim()
     return @reject 'You must provide an email for the user.', 'email', callback
 
-  @saveItem user, callback
\ No newline at end of file
+  @saveItem user, callback
+
+
+exports.sendEmailValidation = (callback) ->
+  @collections.users.exec('sendEmailValidation', userId).pop().callback callback
```
amberwebb: Where does `userId` come from? Does it need to be passed as an argument into the function?

matthopson: Uh, well that's weird. I removed that param, but it must have made it back after a rebase? Good catch.

PR #199, opened 2017-09-29
Linking href check

matthopson:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # TeamSnap JavaScript SDK CHANGELOG
 
+### September 29, 2017 // Version 1.44.0
+- Bug Fix: Check for presence of href during linking.
```
amberwebb: This will be 1.44.1 won't it?

matthopson:
@amberwebb Yeah, I actually looked at the wrong version on that. Good catch.

PR #200, opened 2017-10-09
Reload phone numbers, email addresses and members

amberwebb:
Reloads objects to support combined contacts UI.


src/persistence.coffee
```diff
@@ -289,7 +289,9 @@ modifySDK = (sdk) ->
     (phoneNumber, callback) ->
       saveMemberPhoneNumber.call(this, phoneNumber, callback)
       .then((result) ->
-        sdk.loadMembers({id: phoneNumber.memberId}).then -> result
+        sdk.loadMembers({id: phoneNumber.memberId})
```
matthopson: I think you want to wrap this in `promises.when()`. The `.then -> result` is only being called after `loadContactPhoneNumbers` is finished - so you may end up with a race condition with the `loadMembers` call.

src/persistence.coffee
```diff
@@ -905,6 +907,18 @@ modifySDK = (sdk) ->
           sdk.loadMembers({id: memberId})
           sdk.loadContacts({memberId: memberId})
           sdk.loadContactEmailAddresses({memberId: memberId})
+          sdk.loadContactPhoneNumbers({memberId: memberId})
```
matthopson: Same with these.

src/persistence.coffee
```diff
@@ -905,6 +907,18 @@ modifySDK = (sdk) ->
...
+
+  wrapMethod sdk, 'saveContact',
+    (saveContact) ->
+      (contact, callback) ->
+        saveContact.call(this, contact, callback)
+        .then((result) ->
+          contactId = contact.id
+          sdk.loadMembers({id: contact.memberId})
+          sdk.loadContactEmailAddresses({contactId: contactId})
+          sdk.loadContactPhoneNumbers({contactId: contactId})
```
matthopson: And these. You're not returning the `result` after these finish.

amberwebb:
@matthopson How about now?

PR #202, opened 2017-10-12
OEV-58 Add team id for post validation redirect

kwoodard22:
**Jira Ticket:** https://teamsnap.atlassian.net/browse/OEV-58

Adding `team_id` to the validation link so that we can redirect back to the team's page after successful email validation.

**Classic:** https://github.com/teamsnap/classic/pull/7278
**ApiV3:** https://github.com/teamsnap/apiv3/pull/3295
**Mcfeely:** https://github.com/teamsnap/mcfeely/pull/626
**Cogsworth:** https://github.com/teamsnap/cogsworth/pull/166
**Nextjenn-web:** https://github.com/teamsnap/nextjenn-web/pull/2775
**Teamsnap-javascript-sdk:** https://github.com/teamsnap/teamsnap-javascript-sdk/pull/202


src/collections/users.coffee
```diff
@@ -24,5 +24,5 @@ exports.saveUser = (user, callback) ->
   @saveItem user, callback
 
 
-exports.sendEmailValidation = (callback) ->
-  @collections.users.exec('sendEmailValidation').pop().callback callback
+exports.sendEmailValidation = (params, callback) ->
+  @collections.users.exec('sendEmailValidation', params).pop().callback callback
```
amberwebb: Do you need some validation here? 

For example https://github.com/teamsnap/teamsnap-javascript-sdk/blob/master/src/collections/teamMedia.coffee#L117

The pattern we've followed in these methods is to pass in an item or an id. So I could call `teamsnap.sendEmailValidation(team)` or `teamsnap.sendEmailValidation(teamId)`

kwoodard22:
@amberwebb: Updated! Thanks for pointing to examples as well.

src/collections/users.coffee
```diff
@@ -24,5 +24,9 @@ exports.saveUser = (user, callback) ->
   @saveItem user, callback
 
 
-exports.sendEmailValidation = (callback) ->
-  @collections.users.exec('sendEmailValidation').pop().callback callback
+exports.sendEmailValidation = (teamId, callback) ->
```
matthopson: So, since this endpoint has seen some iteration, I wonder if keeping the more generic `params` object there to hide additional changes would be better. I do personally like the readability of explicit args, which we typically use in most methods, but we do have a few edge cases where this was more of a challenge.

In this case something like this allows you to pass the params object, and validate the params as you go:

```javascript
exports.sendEmailValidation = (params, callback) ->
  unless params?
    throw new TSArgsError 'teamsnap.sendEmailValidation', 'must provide an object with params'
  unless params.teamId? and @isId params.teamId
    throw new TSArgsError 'teamsnap.sendEmailValidation', 'must provide a teamId'
  
  @collections.users.exec('sendEmailValidation', params).pop().callback callback
```

PR #204, opened 2017-10-16
Remove reloading contactEmailAddresses in saveMemberEmailAddress

amberwebb:


matthopson:
Do you need to remove this completely? Or do you want to use the new feature flag?

amberwebb:
@matthopson It should be removed completely. It was added specifically for CCC, but causes side effects.

PR #205, opened 2017-10-16
Support feature flag

matthopson:
- Adds a `features` object to `options` object when authenticating.
- Use `features` for conditional on `invite` in persistence layer.


src/persistence.coffee
```diff
@@ -710,11 +710,14 @@ modifySDK = (sdk) ->
...
           memberId = options.memberId
           promises.when(
-            sdk.loadMemberEmailAddresses({memberId: memberId})
             sdk.loadContactEmailAddresses({memberId: memberId})
-            sdk.loadMemberPhoneNumbers({memberId: memberId})
             sdk.loadContactPhoneNumbers({memberId: memberId})
             sdk.loadMembersPreferences({memberId: memberId})
+
+            # If CCC is enabled, do not reload these collections
+            unless sdk.features.combinedContactCards is true
```
dmccraw: if you don't pass any features during init is `sdk.features` nil?

matthopson: Yeah, I need to add a default. I was passing the default from NJW, which is not the best way to do that.

PR #211, opened 2018-02-15
TS-1643 Reload contacts for member

amberwebb:
When saving a contact, reload all contacts for the member, not just the new contact.

Supports https://github.com/teamsnap/nextjenn-web/pull/2947


src/persistence.coffee
```diff
@@ -927,6 +927,7 @@ modifySDK = (sdk) ->
           contactId = contact.id
           promises.when(
             sdk.loadMembers({id: contact.memberId})
+            sdk.loadContacts({memberId: contact.memberId})
```
matthopson: Do we need to do the same for the emailAddresses and phoneNumbers?

amberwebb: @matthopson Updated. ^^ Thanks.

PR #222, opened 2019-03-11
Fix security vulnerability alerts

VEkh:
Fixes all but two low priority security vulnerabilities that are only related to Brunch (asset builder).

Co-authored-by: Dorian Karter <dorian@hashrocket.com>
Co-authored-by: Vidal Ekechukwu <vidal@hashrocket.com>

PR #233, opened 2020-02-06
Run npm audit fix

bradcypert:
I just simply ran `npm audit fix` on this repo to help fix some issues. I struggled to get tests to fully run before hand, but I am still successfully building the output script files. If someone who's familiar with how these are consumed could drop-in replace their current version's output with these to test and ensure nothing is broken, that'd be awesome.


ryanwi:
good idea, hashrocket did that about a year ago I think to get up to date.  It didn't update the collections files though. Do you know why those all changed here?

https://github.com/teamsnap/teamsnap-javascript-sdk/pull/222/files

bradcypert:
I actually dont know. I'll do some investigation into this and see if i can figure it out.

PR #243, opened 2021-05-04
BAM-2550 add dsp_payload endpoint

bradjteamsnap:
**Story / Task Link**
https://teamsnap.atlassian.net/browse/BAM-2550

### Description

Adding a method for the `dsp_payload` endpoint.

### What did I do?

- Add endpoint to 'users' controller
- Add readme section
- bump package version
- run `npm run build` to generate teamsnap.js and teamsnap.min.js

### How did I test it?

- import remote branch into nextjen-webb
- run console commands with local snapi

### Screenshots

all params
invalid memberId
empty params
memberId


lib/teamsnap.js
```diff
@@ -7616,4 +7627,4 @@ teamsnap.underscoreType = function(type) {
 
 
 global['teamsnap'] = require('teamsnap');
-})();
```
BillyFigueroa, line 7619: I think if there is no end line the linter will throw an issue if there is a linter

BillyFigueroa (approved):

lib/teamsnap.js
```diff
@@ -7471,7 +7482,7 @@ ref = require('./model'), Collection = ref.Collection, Item = ref.Item;
 require('./errors');
 
 TeamSnap = (function() {
-  TeamSnap.prototype.version = '1.61.1';
+  TeamSnap.prototype.version = '1.62.1';
```
Peterortiz82, line 7485: I would probably make this version `1.62.0` instead of `1.62.1`. I believe the [last minor bump](https://github.com/teamsnap/teamsnap-javascript-sdk/commit/5fc3dbe25781a1fc65364ddcd687c328fa7bf17e#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) which took it to a `.1` was for security updates. If we were to tweak or make a minor update to this after release, I'd recommend `1.62.1`.

docs/collections/users.md
```diff
@@ -4,6 +4,8 @@
 
 - [loadUsers](#loadUsers)
 - [saveUser](#saveUser)
+- [sendEmailValidation](#sendEmailValidation)
```
Peterortiz82, line 7: Thanks for adding this!

docs/collections/users.md
```diff
@@ -84,3 +86,27 @@ link provided in the email.
...
 ```
+
+
+---
+
+
+<a id="dspPayload"></a>
+## `dspPayload(params, callback)`
+This is a query endpoint off of user. memberId is the only required param, 
+zone & kuid are optional, but should always be passed in so they will be passed back in the payload.
```
Peterortiz82, line 97: This (zone & kuid) may not apply to the web client. I would check the NextJenn codebase to see if there are any references to Zone or KUID in the code or ask Sean if he has seen these values being passed over by the web client. 

I believe zone represents a screen (event details, postgame report, etc.) within the mobile clients, and KUID may only be iOS specific as it represents the device id. I'm not 100% sure, but I would double-check, and if what I said is true, we can remove this line or explicitly state that these are only currently being used on the mobile clients.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "teamsnap.js",
-  "version": "1.61.1",
+  "version": "1.62.1",
```
Peterortiz82, line 3: same as the above comment changing this to `1.62.0`.

src/teamsnap.coffee
```diff
@@ -3,7 +3,7 @@ promises = require './promises'
 require './errors'
 
 class TeamSnap
-  version: '1.61.1'
+  version: '1.62.1'
```
Peterortiz82, line 6: same here about the version number `1.62.0`

Peterortiz82 (changes requested):
Overall, this looks great! Nice work! Just a small change request to set the version to `1.62.0` instead of `1.62.1` since it's something new that hasn't been updated yet. I would also I'd check to see if zone & KUID applies to the web. If it doesn't, then I would update the `users.md` description for it.

docs/collections/users.md
Peterortiz82: Based on the conversation we had in standup, we use this in the web client, so disregard this comment.

lib/teamsnap.js
bradjteamsnap: Deleted during `npm run build`, I manually added back just in case

Peterortiz82 (approved):
Great job! :shipit:

